PR #4, opened 2021-04-05
Improve search relevance

StewartJingga:
Changes
1. Refactor search module
2. Add query time boosting for ID and Title fields
3. Enable fuzzy search


store/search.go
```diff
@@ -15,68 +16,41 @@ import (
 var (
 	filterScriptBasePredicate       = `doc.containsKey(%q) == false || doc[%q].size() == 0`
 	filterScriptMatchValuePredicate = `doc[%q].value == %q`
+	defaultMaxResults               = 200
+	defaultMinScore                 = 0.01
+	defaultCachedTypesMapDuration   = 300
```
irainia, line 21: Suggestion to put the cache into the config as an option. This way the user can flexibly adjust to their need.

store/testdata/search-test-fixture.json
```diff
@@ -0,0 +1,135 @@
+{
```
irainia, line 1: Suggestion to split the sample data and testable. This is to make it more consistent with the other sample data and to hopefully make it easier to understand the unit test, since separating them will make it more clarity.

PR #21, opened 2021-11-05
fix: encoded urn not matched properly to routes

StewartJingga:


api/middlewares.go
```diff
@@ -0,0 +1,21 @@
...
+	"net/url"
+
+	"github.com/gorilla/mux"
+)
+
+func decodeURLMiddleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		newVars := map[string]string{}
+		for key, val := range mux.Vars(r) {
+			decodedValue, _ := url.QueryUnescape(val)
```
mabdh, line 14: Should we log error here instead of ignoring it?

StewartJingga: we could, but the log wouldn't get called ever.

checking from the [doc](https://pkg.go.dev/net/url#QueryUnescape), `QueryUnescape` will return error `"if any % is not followed by two hexadecimal digits."`.

and if somehow client sends an invalid encoded url, it would not even reach this middleware because somehow gorilla.mux already returning 400 status code by default.

and that's why I skipped it. Although  🤔 , if somehow gorilla.mux changes its behaviour we might need to log it.
wdyt?

StewartJingga: @mabdh I updated the code to handle if gorilla mux changes its behaviour 👌

mabdh: Oh yes cool. The middleware could have a logger param.

mabdh (approved):

PR #29, opened 2021-11-29
feat(tag): add record tagging feature

StewartJingga:


tag/sqlstore/tag_repository.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	recordModelTemplate, err := r.getTemplateWithFields(domainTag.TemplateURN)
+	if err != nil {
+		return err
+	}
+	domainTemplate := r.convertToDomainTemplate(recordModelTemplate)
+	err = r.dbClient.Transaction(func(tx *gorm.DB) error {
+		return r.insertDomainTagToDB(tx, domainTag)
+	})
+	if err == nil {
```
mabdh, line 40: We could check err! =nil first here right? To avoid nested if

tag/template_service.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	filterForExistence := Template{
+		URN: template.URN,
+	}
+	templateRecords, err := s.repository.Read(filterForExistence)
+	if err != nil {
+		return errors.Wrap(err, "error checking template existence")
+	}
+	if len(templateRecords) > 0 {
```
mabdh, line 51: If Read always expect 1 result, then why not doing this validation inside the read itself?

tag/template_service.go
```diff
@@ -0,0 +1,158 @@
...
+		return err
+	}
+	filterForExistence := Template{
+		URN: template.URN,
+	}
+	templateRecords, err := s.repository.Read(filterForExistence)
+	if err != nil {
+		return errors.Wrap(err, "error checking template existence")
+	}
+	if len(templateRecords) == 0 {
```
mabdh, line 88: I guess this should be done in Read repository layer

tag/template_service.go
```diff
@@ -0,0 +1,158 @@
...
+		return TemplateNotFoundError{URN: template.URN}
+	}
+
+	// check for duplication
+	templateFromDB := templateRecords[0]
```
mabdh, line 93: Same with this, if template supposed to be unique with its urn, we could agree that Read should always return 1 result. So we  do this in repository layer instead.

tag/template_service.go
```diff
@@ -0,0 +1,158 @@
...
+func (s *TemplateService) Find(urn string) (Template, error) {
+	queryDomainTemplate := Template{
+		URN: urn,
+	}
+	listOfDomainTemplate, err := s.repository.Read(queryDomainTemplate)
+	if err != nil {
+		return Template{}, errors.Wrap(err, "error reading repository")
+	}
+	if len(listOfDomainTemplate) == 0 {
+		return Template{}, TemplateNotFoundError{URN: urn}
```
mabdh, line 132: This is also the same. Checking this supposed to be the responsibility of repository layer.

tag/template_service.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error deleting template")
+	}
+	return nil
+}
+
+// NewTemplateService initializes service template service
+func NewTemplateService(r TemplateRepository) *TemplateService {
+	if r == nil {
+		panic("repository is nil")
```
mabdh, line 152: Better to not have panic inside logic or inner layers. Panic supposed to be happened in cmd or main file. Better to return error here and panic out in the initialization if error.

tag/validator.go
```diff
@@ -0,0 +1,103 @@
...
+
+// newTemplateValidator initializes validator for tag template
+func newTemplateValidator() validator.Validator {
+	v, err := validator.NewBuilder().
+		WithStructValidations([]validator.StructValidation{
+			{
+				Type: Template{},
+				Func: func(sl v.StructLevel) {
+					template := sl.Current().Interface().(Template)
+					for i, field := range template.Fields {
```
mabdh, line 46: This could lead to nil pointer panic if the interface above failed to be casted. Better to check if casting is ok or not.

tag/validator/builder.go
```diff
@@ -0,0 +1,169 @@
...
+
+		errMsg := f.Translate(b.translator)
+		fieldErrors[field] = errMsg
+	}
+	return fieldErrors
+}
+
+func (b *Builder) initializeTranslator() ut.Translator {
+	universalTranslator := ut.New(en.New(), en.New())
+	translator, _ := universalTranslator.GetTranslator(defaultLocale)
```
mabdh, line 103: Is the second return param an error? Why do we ignore it?

api/handlers/tag_template_handler.go
```diff
@@ -0,0 +1,146 @@
...
+		return
+	}
+	err := h.service.Create(&requestBody)
+	if errors.As(err, new(tag.DuplicateTemplateError)) {
+		writeJSONError(w, http.StatusConflict, err.Error())
+		return
+	}
+	if err != nil {
+		internalServerError(w, h.logger, fmt.Sprintf("error creating tag template: %s", err.Error()))
+
```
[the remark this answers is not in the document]
StewartJingga: yeah it is already been formatted. nothing changes after running gofmt 🤔 .
will remove this manually.

tag/sqlstore/tag_repository.go
StewartJingga: yeah I guess we could, the logic is directly from dexter, if the test is not breaking will apply the changes.

tag/template_service.go
StewartJingga: checking from Dexter code, it seems like they design `Read` to return list.
It is also being used to fetch list of templates in this file on `Index` method.

tag/template_service.go
StewartJingga: Read currently returns list of templates, and it is also being used for fetching list of templates by TemplateService.

tag/template_service.go
StewartJingga: Agree, ideally template repository could have a method to fetch a single record. But I decided to do major changing on Dexter code signature to avoid pinballing code changes after another.

tag/template_service.go
StewartJingga: yeah either that or i'm actually thinking to remove this repository nil check in this part 🤔 . wdyt?

tag/validator/builder.go
StewartJingga: I don't meddle much in this code since it is a working code on Dexter so far, but I will try to handle the error and see if something breaks.

tag/validator.go
StewartJingga: yeah same, didn't want to change Dexter logic before, but will try to see if we can do that.

tag/validator/builder.go
StewartJingga: apparently the second param is not an error but a `found bool` instead. will ignore it for now since I'm still not sure how to handle it.

tag/validator.go
StewartJingga: updated

tag/template_service.go
mabdh: Considering this will only be called in the initialization, it is fair enough to remove this.

mabdh (approved):

PR #37, opened 2021-12-08
feat(discovery): boost search result with table usage

mabdh:
Changes
- Add a new query param `sortby` in the API that accepts `usage` (only for now). If this is passed, we return search results that sorted by table usage (the bigger usage_count, the more it prioritized to appear first)
- Convert ES query
From this
```
{
	"query": {
		"bool": {
			"should": [
				{
					"multi_match": {
						"fields": [
							"urn^10",
							"name^5"
						],
						"query": "sample"
					}
				},
				{
					"multi_match": {
						"fields": [
							"urn^10",
							"name^5"
						],
						"fuzziness": "AUTO",
						"query": "sample"
					}
				},
				{
					"multi_match": {
						"fuzziness": "AUTO",
						"query": "sample"
					}
				}
			]
		}
	},
	"min_score": 0.01
}
```

To use `function score query`
```
{
	"query": {
		"function_score": {
			"functions": [
				{
					"field_value_factor": {
						"field": "data.profile.usage_count",
						"missing": 1,
						"modifier": "log1p"
					},
					"weight": 1
				}
			],
			"query": {
				"bool": {
					"should": [
						{
							"multi_match": {
								"fields": [
									"urn^10",
									"name^5"
								],
								"query": "sample"
							}
						},
						{
							"multi_match": {
								"fields": [
									"urn^10",
									"name^5"
								],
								"fuzziness": "AUTO",
								"query": "sample"
							}
						},
						{
							"multi_match": {
								"fuzziness": "AUTO",
								"query": "sample"
							}
						}
					]
				}
			},
			"score_mode": "sum"
		}
	},
	"min_score": 0.01
}
```

### DIFF
--- a/api/handlers/search_handler.go
+++ b/api/handlers/search_handler.go
@@ -63,6 +63,7 @@ func (handler *SearchHandler) buildSearchCfg(params url.Values) (cfg discovery.S
 	cfg.Text = text
 	cfg.MaxResults, _ = strconv.Atoi(params.Get("size"))
 	cfg.Filters = filterConfigFromValues(params)
+	cfg.SortBy = params.Get("sortby")
 	cfg.TypeWhiteList, err = parseTypeWhiteList(params)
 	return
 }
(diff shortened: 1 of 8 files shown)


StewartJingga:
instead of hardcoding `data.profile.usage_count`, could we get it from querystring so that users can sort using another field? This is just to make it more generic not specific to our use case because `data.profile.usage_count` is our model while users are free to use any structure that they want.

wdyt? @mabdh

mabdh:
> instead of hardcoding `data.profile.usage_count`, could we get it from querystring so that users can sort using another field? This is just to make it more generic not specific to our use case because `data.profile.usage_count` is our model while users are free to use any structure that they want.
> 
> wdyt? @mabdh

yes, that could be done. let me update the code.

PR #41, opened 2021-12-14
feat(discovery): search by record's field name

mabdh:


record/record.go
```diff
@@ -25,6 +26,12 @@ type LineageRecord struct {
 	Type string `json:"type"`
 }
 
+type Owner struct {
+	URN  string `json:"urn"`
```
StewartJingga, line 30: I'm thinking if we should have an email field too 🤔 .
There might be a case where URN won't be an email. It could be an ID or even username e.g. `john.doe`.

wdyt? @mabdh @ravisuhag

mabdh: For now, would it be a problem if `Name` could contain `e-mail` or `Full name` or any specific identifier of the owner?
The different between `urn` and `name` should be just `name` could be non-unique while `urn` should be unique across our platform context.

ravisuhag: Putting email in the name could be misleading. If we want to differentiate then email is better I would say.

mabdh: Make sense, I will update assets schema and add a new field called email in this `Owner` struct. I will also update the `Owner` in meteor sink. (Although in the future we could just use columbus package to do that.

@StewartJingga in that case we need to update the extractor too.

mabdh: Created

https://github.com/odpf/proton/pull/80

https://github.com/odpf/meteor/issues/283

store/elasticsearch/search.go
```diff
@@ -203,7 +203,35 @@ func (sr *Searcher) buildSuggestQuery(ctx context.Context, cfg discovery.SearchC
 	return payload, err
 }
 
-func (sr *Searcher) buildTextQuery(ctx context.Context, text string) elastic.Query {
+func (sr *Searcher) buildTextQuery(ctx context.Context, cfg discovery.SearchConfig) elastic.Query {
+	if len(cfg.Queries) == 0 {
+		return sr.buildTextQueryGeneric(ctx, cfg.Text)
```
StewartJingga, line 208: shouldn't `buildTextQuery` will always run regardless of `cfg.Queries`?
so that `cfg.Queries` will only add another filter query on top of existing query.

store/elasticsearch/search.go
```diff
@@ -246,6 +274,7 @@ func (sr *Searcher) buildFilterQueries(query elastic.Query, filters map[string][
 			values = append(values, rawVal)
 		}
 
+		key := fmt.Sprintf("%s.keyword", key)
```
StewartJingga, line 277: this means field without `keyword` will not get matched?

store/elasticsearch/search.go
mabdh: Since we are using several multimatch queries and boosting urn and name field, the cfg Queries search will take a little impact. Even if we boost the cfg Queries field itself. The behaviour is not as expected.

store/elasticsearch/search.go
mabdh: yes, this should be consistent with filtering in get all
https://github.com/odpf/columbus/blob/main/store/elasticsearch/record_repository.go#L208

store/elasticsearch/search.go
mabdh: But if it is `filter`, that supposed to be part of cfg.Filters right?

StewartJingga (approved):
LGTM

PR #51, opened 2022-01-06
refactor(types): clean up dynamic types

mabdh:
**Tasks**
- Add Elasticsearch migration to migrate command to create indices for all of our available types
- Change Type to enum
- Remove types CRUD APIs


record/type.go
```diff
@@ -44,19 +64,39 @@ var AllTypeClassifications = []TypeClassification{
 // Entities are supposed to represent resources, datasets and schema.
 // XXX(Aman): should Type names be case insensitive?
 type Type struct {
-	Name           string             `json:"name"`
+	Name           TypeName           `json:"name"`
 	Classification TypeClassification `json:"classification"`
```
StewartJingga, line 68: CMIIW but is the reason you keep this Type struct because of `classification` field?
We can actually remove `classification` field since we are not using it anywhere else now.

cmd/migrate.go
```diff
@@ -22,17 +46,35 @@ func Migrate() {
...
+	log.Info("Initiating ES client...")
+	esClient := initElasticsearch(config)
+	tr := esStore.NewTypeRepository(esClient)
+	for _, supportedType := range record.AllSupportedTypes {
+		log.Infof("Migrating %q type\n", supportedType.Name)
+		ctx, cancel := context.WithTimeout(context.Background(), esMigrationTimeout)
+		defer cancel()
+		err = tr.CreateOrReplace(ctx, supportedType)
+		if err != nil {
+			err = fmt.Errorf("error creating/replacing type: %q, err: %w", supportedType.Name, err)
```
StewartJingga, line 74: we can change this to 
```suggestion
			err = errors.Wrapf(err, "error creating/replacing type: %q", supportedType.Name)
```

record/type.go
mabdh: Tbh because I don't know it is important or not 😂. Ok thanks stew, will remove it.

store/elasticsearch/type_repository.go
```diff
@@ -39,32 +39,32 @@ type TypeRepository struct {
 	cli *elasticsearch.Client
 }
 
-func (repo *TypeRepository) CreateOrReplace(ctx context.Context, recordType record.Type) error {
-	if isReservedName(recordType.Name) {
-		return record.ErrReservedTypeName{TypeName: recordType.Name}
+func (repo *TypeRepository) CreateOrReplace(ctx context.Context, recordTypeName record.TypeName) error {
```
StewartJingga, line 42: We may not need this function anymore if we are going with type as an enum.

record/type.go
```diff
@@ -21,48 +20,40 @@ type TypeFields struct {
...
 }
 
-// Type represents a named collection of records
-// Entities are supposed to represent resources, datasets and schema.
-// XXX(Aman): should Type names be case insensitive?
+// Type represents a typename wrapped in a JSON
 type Type struct {
-	Name           string             `json:"name"`
-	Classification TypeClassification `json:"classification"`
+	Name TypeName `json:"name"`
```
StewartJingga, line 40: Do we still need Type struct or we can operate with enum only?

store/elasticsearch/type_repository.go
StewartJingga: In fact, is it possible to remove TypeRepository? Or do we still need it for something else?

mabdh: We need this for migration.

record/type.go
mabdh: Unless we want to get rid of meta index, we need this type struct to add a new type in meta index.

StewartJingga: we can get rid of `_meta` index, we will not use it anymore.

store/elasticsearch/type_repository.go
StewartJingga: Oh yeah true!
Would it make sense if we move `record.TypeRepository.CreateOrReplace` to `discovery.Migrate(typeName)`?
wdyt?

mabdh: hmm I think better to separate the responsibility to repository layer

StewartJingga: Oh true forgot about it 🙏.  I mean `store/elasticsearch.Migrate(typeName)`.

mabdh: Hmm yes that might be doable. Will try it.

mabdh: @StewartJingga I think better to just keep it in typeRepository, because other functions need that to test

StewartJingga (approved):
LGTM

PR #53, opened 2022-01-09
fix: remove build action workflow

StewartJingga:
Fix github's `build` action workflow due to code changes made to cater new `release`action workflow.


.github/workflows/build.yml
```diff
@@ -1,28 +1,37 @@
 name: Build
+
 on:
   push:
-    tags:
-      - "v*.*.*"
+    branches:
```
ravisuhag, line 5: @StewartJingga Any reason we want to release docker image on every commit?

StewartJingga: Was just letting people to get the latest update/fix on main branch only.

Right now `latest` tag in docker means latest main branch. unless we want to change it to mean latest tag.

ravisuhag: I think we should point the latest docker tag to the latest stable release and we can push a nightly build with a different tag for the main branch build. WDYT?

StewartJingga: sounds good to me. after reading a few articles about latest tag, it seems like it is to be avoided if possible.
latest is basically just another tag that is being added if there is no tag presents.

So maybe we could just remove this `build` action unless we want a special tag for this main branch called `main` perhaps? wdyt?

ravisuhag: For now, I think we can remove it. If later we decide to publish the nightly build we can include it.

StewartJingga: @ravisuhag removed 🙏

PR #61, opened 2022-01-12
refactor(api): migrate v1 to v1beta1

mabdh:
- Instead of removing `v1` base URL, I am duplicating the API to `v1beta1`
- Will remove the `v1` in the near future


StewartJingga (approved):
Code's LGTM, but need to check on integration whether newrelic catch the routes properly.
There was an issue before regarding handlers not being set on root router.

PR #65, opened 2022-01-19
refactor: replace gorm with jmoiron/sqlx 

mabdh:


store/postgres/model.go
```diff
@@ -1,40 +1,288 @@
...
+	RecordType string    `db:"record_type"`
+	FieldID    uint      `db:"field_id"`
+	CreatedAt  time.Time `db:"created_at"`
+	UpdatedAt  time.Time `db:"updated_at"`
+	Field      Field     `db:"-"`
+}
+
+type Tags []Tag
+
+func (ts Tags) buildMapByTemplateURN() map[string][]Tag {
```
mabdh, line 24: note: might need to add more tests to these struct methods

migrations/000001_create_tags_templates_fields_table.up.sql
```diff
@@ -0,0 +1,38 @@
+BEGIN;
```
ravisuhag, line 1: @mabdh Since these migrations are specific to postgres. They should better be placed in store/postgres package only.

mabdh: Yes it does make sense to store it inside postgres libraries. Will update it.

mabdh: @ravisuhag I have moved the folder and also update the migration with `go:embed` instead of source files directly.

store/postgres/migrations/000001_create_tags_templates_fields_table.down.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+DROP TABLE IF EXISTS tags;
+
+DROP TABLE IF EXISTS fields;
+
+DROP TABLE IF EXISTS templates;
+
```
rahmatrhd, line 8: shouldn't we drop the indexes as well?

mabdh: AFAIK all constraints (including indices) are also get removed once we drop the table.

store/postgres/migrations/000001_create_tags_templates_fields_table.up.sql
```diff
@@ -0,0 +1,38 @@
+BEGIN;
+
+CREATE TABLE templates (
```
StewartJingga, line 3: let's rename this to `tag_templates` to make sure that this is for tagging since template is pretty general and  Columbus might have other use case for template later on.

store/postgres/migrations/000001_create_tags_templates_fields_table.up.sql
```diff
@@ -0,0 +1,38 @@
...
+    urn text PRIMARY KEY,
+    display_name text NOT NULL,
+    description text NOT NULL,
+    created_at timestamp DEFAULT NOW(),
+    updated_at timestamp DEFAULT NOW()
+);
+
+CREATE TABLE fields (
```
StewartJingga, line 11: same goes with this, we could rename this to `tag_template_fields`.

cmd/serve.go
```diff
@@ -80,11 +79,16 @@ func initRouter(
...
 
-	pgClient := initPostgres(config)
-	tagRepository := postgres.NewTagRepository(pgClient)
-	tagTemplateService := tag.NewTemplateService(
-		postgres.NewTemplateRepository(pgClient),
-	)
+	pgClient := initPostgres(rootLogger.WithField("reporter", "postgres"), config)
+	tagRepository, err := postgres.NewTagRepository(pgClient)
+	if err != nil {
+		log.Fatal(err)
```
StewartJingga, line 85: should we somehow format/wrap the error here for easier debugging?
e.g. `error initiating tag repository: %w`

StewartJingga (changes requested):

tag/errors.go
```diff
@@ -4,13 +4,13 @@ import (
 	"fmt"
 )
 
-type NotFoundError struct {
+type ErrNotFound struct {
```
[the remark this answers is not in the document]
mabdh: There is actually no specific consideration for this.

I was just feeling putting all error into the same context (having `Err` prefix) would be better for readability. When we call a package, we know all possible errors are Within `Err*`. 

It is also okay to revert it back to follow what written in the doc.

store/postgres/migrations/000001_create_tags_templates_fields_table.up.sql
mabdh: Sure, but in that case, I think we should also rename and do some refactor in the repository model too.

cmd/serve.go
mabdh: yeah, can do that.

tag/errors.go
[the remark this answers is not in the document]
StewartJingga: > When we call a package, we know all possible errors are Within Err*.

Ah I see, makes sense. But I think let's differentiate error struct with error variable, I even see golang core `os` package use the `*Error` convention for its struct here. 🙏 
https://pkg.go.dev/os#LinkError

StewartJingga (approved):

ravisuhag (approved):

PR #66, opened 2022-01-19
refactor: inject logger from the top in the required packages

scortier:


api/handlers/record_handler.go
```diff
@@ -89,9 +88,7 @@ func (h *RecordHandler) UpsertBulk(w http.ResponseWriter, r *http.Request) {
 	var failedRecords = make(map[int]string)
 	for idx, record := range records {
 		if err := h.validateRecord(record); err != nil {
-			h.logger.WithField("type", typName).
-				WithField("record", record).
-				Errorf("error validating record: %v", err)
+			h.logger.Error("error validating record", "type", typName, "record", record, "error", err)
```
StewartJingga, line 91: @scortier if we print the error without formatting it `e.g. using %v`, will it still log the same value like before?

scortier: Yes, it will log the same value, have used it in Optimus too

StewartJingga (approved):

PR #67, opened 2022-01-21
feat: add user repository

mabdh:
rebased on https://github.com/odpf/columbus/pull/65

will assign to be reviewed once https://github.com/odpf/columbus/pull/65 has got merged.


user/service.go
```diff
@@ -0,0 +1,25 @@
+package user
+
+import "context"
+
+// Service is a type of service that manages business process
+type Service struct {
```
StewartJingga, line 6: If there is no business logic, should we skip Service struct and go straight to UserRepo? wdyt?
We can always introduce service layer when business logic arises.

Im still thinking if it is a good idea to create service just for the sake of having a service. 🤔

user/service.go
```diff
@@ -0,0 +1,25 @@
...
+	repository Repository
+}
+
+// Create handles create business operation for user
```
StewartJingga, line 10: wrong comment, might've missed it.

StewartJingga (changes requested):

user/service.go
mabdh: agree with this

user/service.go
mabdh: this will be removed in this PR

StewartJingga (approved):
LGTM

Issue #39, opened 2021-12-10
feat(lint): add `golangci-lint` to the workflow


mabdh:
**Is your feature request related to a problem? Please describe.**
This is related with issue #38. Once we fixed all lint warning/error, we need to add a `lint` workflow to automate lint checking.

**Describe the solution you'd like**
Use `golangci-lint` with default [linters](https://golangci-lint.run/usage/linters/)

Issue #38, opened 2021-12-10
fix(lint): fix all linter warning


mabdh:
**Is your feature request related to a problem? Please describe.**
When doing lint with `golangci-lint`. I found there are several linter warning. We could fix all warning to avoid some unwanted bugs or unintended behaviour. After fixing all lint warning/error, we could add a new `lint` in github workflow.

**Describe the solution you'd like**
Fixing every line of code that throw lint warning/error when using `golangci-lint`.

PR #69, opened 2022-01-24
fix(lint): fix linting error

scortier:
- Add `golangci-lint` to the workflow
- Add `lint` command 
- Resolve linting errors

### DIFF
--- a/lineage/graph.go
+++ b/lineage/graph.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/odpf/columbus/lib/set"
 	"github.com/odpf/columbus/record"
-	"github.com/pkg/errors"
 )
 
 type QueryCfg struct {
@@ -32,7 +31,6 @@ func (graph *InMemoryGraph) init() {
 		}
 		values.Add(id)
 	}
-	return
 }
 
 func (graph InMemoryGraph) Query(cfg QueryCfg) (AdjacencyMap, error) {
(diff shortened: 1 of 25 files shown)


scortier:
Related Issues: #38 #39

lineage/graph.go
```diff
@@ -59,7 +57,7 @@ func (graph InMemoryGraph) Query(cfg QueryCfg) (AdjacencyMap, error) {
 		var err error
 		supergraph, err = graph.buildSubgraphFromRoot(supergraph, cfg.Root)
 		if err != nil {
-			return supergraph, errors.Wrap(err, "error building subgraph")
+			return supergraph, fmt.Errorf("error building subgraph %s", err)
```
mabdh, line 60:
```suggestion
			return supergraph, fmt.Errorf("error building subgraph: %w", err)
```
If it is error wrapping, we could use `%w`.
see https://go.dev/blog/go1.13-errors

scortier: Should we replace all error wrapping with `fmt.Errorf` as `errors.wrap` package has been archived.
wdyt ??

mabdh: yes let's do it. I guess that should be no problem, we could do that as part of this story.

scortier: sounds good

mabdh:
@scortier I still see `github.com/pkg/errors` package in `go.mod`. Have you tried doing `go mod tidy`?

scortier:
> @scortier I still see `github.com/pkg/errors` package in `go.mod`. Have you tried doing `go mod tidy`?

Yes, I did

mabdh (approved):

PR #70, opened 2022-01-25
feat: update api to be user-aware

mabdh:
Currently require all APIs in `v1beta1` to have identity headers (configurable via config)
- `Columbus-User-Email` ex: `user@odpf.io`

`v1` API (that we want to remove) won't need that header


store/postgres/tag_repository_test.go
```diff
@@ -26,21 +26,21 @@ type TagRepositoryTestSuite struct {
 func (r *TagRepositoryTestSuite) SetupSuite() {
 	var err error
 
-	logger := logrus.New()
+	logger := log.NewLogrus()
```
StewartJingga, line 29: there is `log.NewNoop` from salt to stub logger.

user/service.go
```diff
@@ -0,0 +1,78 @@
...
+// Service is a type of service that manages business process
+type Service struct {
+	repository Repository
+	config     Config
+}
+
+// ValidateUser check if user information is in header
+// if exist in header, validate with user info from DB, if not return error
+// if exist in DB, return user ID, if not exist in DB, create a new one
+func (s *Service) ValidateWithHeader(ctx context.Context, header http.Header) (string, error) {
```
StewartJingga, line 26: is it possible to move this validation on handlers layer 🤔 .
Not sure if we want to user service know http context.
This could impact for when we are introducing CLI interface.

store/postgres/tag_repository_test.go
mabdh: this logger was needed to debug with dockertest. But sure we can use Noop too.

user/service.go
mabdh: sure, let me check

api/middleware/user.go
```diff
@@ -0,0 +1,34 @@
...
+// ValidateUser middleware will propagate a valid user ID as string
+// within request context
+// use `user.FromContext` function to get the user ID string
+func ValidateUser(logger log.Logger, userSvc *user.Service) mux.MiddlewareFunc {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+			userID, err := userSvc.ValidateWithHeader(r.Context(), r.Header)
+			if err != nil {
+				logger.Warn("error validating user", "error", err)
+				if errors.Is(errors.Unwrap(err), user.ErrNoUserInformation) {
```
rahmatrhd, line 31: `errors.Is` handles unwrapping the errors chain so we don't have to unwrap it manually here

api/middleware/user.go
```diff
@@ -0,0 +1,34 @@
...
+					handlers.WriteJSONError(rw, http.StatusBadRequest, err.Error())
```
rahmatrhd, line 32: are we returning `http.StatusBadRequest` because the user info is used for enriching the data/record instead of authorization?

store/postgres/user_model.go
```diff
@@ -6,15 +6,15 @@ import (
 	"github.com/odpf/columbus/user"
 )
 
-type User struct {
+type UserModel struct {
 	ID        string    `db:"id"`
 	Email     string    `db:"email"`
 	Provider  string    `db:"provider"`
```
rahmatrhd, line 12: I haven't understood the context completely, but if the information we're going to store is only email & provider. Since both email & provider are required in all APIs' headers, then why do we need to store it?

api/middleware/user.go
mabdh: yes this is already updated in the latest changes.

store/postgres/user_model.go
StewartJingga: the reason we store email is because all of our features will be using `user_id` for reference.
Since we only receive email from shield, we need to map it to Columbus user's id.

api/middleware/user.go
mabdh: yes, this is not really an authorization, unless we don't do a `real` authorization check, i don't think we should return 401

store/postgres/tag_repository_test.go
StewartJingga: Oh if it is needed then it's fine 👌 . Was just an fyi.

store/postgres/user_model.go
mabdh: email is just the `external` user id (we could consider it as user name). The `id` (uuid) in `user` table is internal user id that we are going to use inside columbus.

rahmatrhd: or something like `account_id`, it should be more generic

mabdh: we were also thinking that way, having a more generic term for external user id. But in that case we can't guarantee the external id would be unique. On the other hand, `email` should be unique and we can make it required as for now.

rahmatrhd: we can create a unique constraint for provider+account_id. But yeah, we can assess that option later once we have more specific use case. Maybe `email` is sufficient for now.

mabdh: yes, we can do that, but currently the `provider` field is not really that required. By default we make it `shield`. 
alright, thanks @rahmatrhd for the input.

store/postgres/tag_repository_test.go
mabdh: no worries, thanks for pointing out stew, I have just updated it to Noop in PR, we can easily make it `NewLogrus` if we want to debug it though.

PR #72, opened 2022-01-25
feat: create Asset model to replace Record

StewartJingga:


asset/asset.go
```diff
@@ -0,0 +1,38 @@
+package asset
+
```
mabdh, line 2: We can add one line here
```suggestion
//go:generate mockery --name Repository --outpkg mock --output ../lib/mock/ --structname AssetRepository --filename asset_repository.go
```

so we can easily generate mock to `lib/mock`

asset/asset.go
```diff
@@ -0,0 +1,38 @@
...
+	GetByID(ctx context.Context, id string) (Asset, error)
+	Upsert(context.Context, *Asset) error
+	Delete(ctx context.Context, id string) error
+}
+
+type GetConfig struct {
+	Text    string `json:"text"`
+	Type    Type   `json:"type"`
+	Service string `json:"service"`
+	Size    int    `json:"size"`
```
mabdh, line 36: I think we need to define a standard for the pagination params. Is it going to be the standard?

store/postgres/asset_model.go
```diff
@@ -0,0 +1,118 @@
...
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/odpf/columbus/asset"
+)
+
+type Asset struct {
```
mabdh, line 13: I propose to standardize the naming of domain and DB model.
Model naming always end with suffix `..Model`
Domain name is the entity name itself.

For example:
The DB model of asset is `AssetModel`
The domain model of asset is `Asset`

To simplify, I think we don't need to explicitly mention that a struct is a `Domain` and we just need to consider every entity packages outside `store` package is our domain codes. Therefore, the function helper to convert DB model to domain could be more explicit. Instead of having a function called `toDomain` or  `toAssetDomain`, we could just have an explicit function `toAsset` where we already know that `Asset` is a domain.

wdyt @StewartJingga ?

store/postgres/asset_repository.go
```diff
@@ -0,0 +1,190 @@
...
+// Upsert creates a new asset if it does not exist yet.
+// It updates if asset does exist.
+// Checking existance is done using "urn", "type", and "service" fields.
+func (r *AssetRepository) Upsert(ctx context.Context, ast *asset.Asset) error {
+	assetID, err := r.getID(ctx, ast)
+	if err != nil {
+		return fmt.Errorf("error getting asset ID: %w", err)
+	}
+	if assetID == "" {
+		assetID, err = r.insert(ctx, ast)
```
mabdh, line 66: any reason why not directly using `INSERT ... ON CONFLICT UPDATE ...` ?

store/postgres/asset_model.go
StewartJingga: Yeah agree with that, never in fond of using domain anyway.

Will apply the changes.

api/handlers/asset_handler.go
```diff
@@ -0,0 +1,188 @@
...
+		assetRepository: assetRepository,
+		discoveryRepo:   discoveryRepo,
+		logger:          logger,
+	}
+
+	return handler
+}
+
+func (h *AssetHandler) Get(w http.ResponseWriter, r *http.Request) {
+	config := h.buildGetConfig(r.URL.Query())
```
mabdh, line 38:
```suggestion
	config := h.buildAssetConfig(r.URL.Query())
```
maybe?

store/postgres/user_repository.go
```diff
@@ -16,14 +17,42 @@ type UserRepository struct {
 
 // Create insert a user to the database
 func (r *UserRepository) Create(ctx context.Context, ud *user.User) (string, error) {
+	return r.create(ctx, r.client.db, ud)
+}
+
+// Create insert a user to the database using given transaction as client
+func (r *UserRepository) CreateWithTx(ctx context.Context, tx *sqlx.Tx, ud *user.User) (string, error) {
```
mabdh, line 24: Our postgres client has a method `RunWithTx`. It is in `postgres.go`, maybe we can use that instead?

mabdh: oh I just saw this is actually a different kind of create. I thought this has several queries inside. I think it is fine for this.

store/postgres/asset_repository.go
```diff
@@ -0,0 +1,401 @@
...
+		return fmt.Errorf("error getting affected rows: %w", err)
+	}
+	if affectedRows == 0 {
+		return asset.NotFoundError{AssetID: id}
+	}
+
+	return nil
+}
+
+func (r *AssetRepository) buildGetQuery(config asset.Config) (query string, args []interface{}) {
```
mabdh, line 123: maybe using `https://github.com/Masterminds/squirrel` would make it more readable?

I was trying using this before with tag template story but the query doesn't really that complex like this one stew.
@StewartJingga

store/postgres/asset_repository.go
```diff
@@ -0,0 +1,401 @@
...
+
+		return nil
+	})
+
+	return
+}
+
+func (r *AssetRepository) update(ctx context.Context, id string, ast *asset.Asset) error {
+	return r.client.RunWithinTx(ctx, func(tx *sqlx.Tx) error {
+		_, err := r.client.db.ExecContext(ctx,
```
mabdh, line 217: should we check `res.RowsAffected()` here instead of ignoring it as `_`?

store/postgres/asset_repository.go
```diff
@@ -0,0 +1,401 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/odpf/columbus/asset"
+	"github.com/odpf/columbus/user"
+)
+
+const (
+	DEFAULT_MAX_SIZE = 100
```
mabdh, line 16:
```suggestion
	MAX_RESULT_SIZE = 100
	MAX_DEFAULT_SIZE = 30
```

I think we should have the number above
MAX_RESULT_SIZE -> to handle if user input size > MAX_RESULT_SIZE, we limit it to MAX_RESULT_SIZE
MAX_DEFAULT_SIZE -> if user doesn't put the size, so checking it will be if size < 1

And this could go to `postgres.go` file so other repository could use the same value.

store/postgres/asset_model.go
```diff
@@ -0,0 +1,118 @@
...
+func (m JSONMap) MarshalJSON() ([]byte, error) {
+	if m == nil {
+		return []byte("null"), nil
+	}
+	t := (map[string]interface{})(m)
+	return json.Marshal(t)
+}
+
+// UnmarshalJSON to deserialize []byte
+func (m *JSONMap) UnmarshalJSON(b []byte) error {
```
mabdh, line 113: Hmm having different kind of function pointer (one pass by reference, other pass by value) looks not inconsistent here. But I can't really think any better portable solution for this 🤔.

store/postgres/asset_repository.go
StewartJingga: hmm do you think we should we limit maximum result size for now?

And for the default max size, `MAX_DEFAULT_SIZE` has different meaning.
It means more to `The maximum value for default size`.

On the other hand, `DEFAULT_MAX_SIZE` means `The default value for max size`. This is what you aim for right with the constant?

store/postgres/asset_model.go
StewartJingga: Yeah, but actually I have seen it a few times, when you want to override behaviour, it is expected I guess.

I actually found this approach from these two articles:
https://medium.com/iostrap/manage-postgresql-json-data-with-go-golang-b0ae416972c5

store/postgres/asset_repository.go
mabdh: Oh yes sorry, what I meant is.

```suggestion
	DEFAULT_MAX_RESULT_SIZE = 100
	DEFAULT_RESULT_SIZE = 30
```
My concern is, there will be a plenty of assets. If our API is being used internally, it is not necessary I think to limit it. But if it will be used externally (assuming there is bad actors), not limiting the result size could impact the performance.

Hmm, let's just use `The default value for max size` for now. But instead of calling it `DEFAULT_MAX_SIZE`, can we make the name clearer by using `DEFAULT_MAX_RESULT_SIZE`.
We can

store/postgres/asset_model.go
mabdh: Hmm yes, it is actually not about the json itself, more into the receiver functions. But I think we can keep this as it is.

store/postgres/asset_repository.go
StewartJingga: nice library 👌

store/postgres/asset_model.go
mabdh: @StewartJingga should we update it to `AssetModel` stew?

StewartJingga: Oh sorry missed this, updating

asset/asset.go
StewartJingga: @mabdh I tried adding the line and run `mockery --all`, but it does not seem to have any difference. Did I miss something?

mabdh: Oh for this, you just need to run `go generate` inside the folder. Or in VScode, there should be a click button helper above it.
Screen Shot 2022-02-02 at 15 09 03

StewartJingga: ah got it, thanks!

mabdh (approved):

PR #73, opened 2022-01-26
feat: add user/starred api

mabdh:


asset/asset.go
```diff
@@ -23,11 +23,24 @@ type Asset struct {
...
+
+	return nil
+}
+
 type Repository interface {
 	Get(context.Context, Config) ([]Asset, error)
 	GetCount(context.Context, Config) (int, error)
 	GetByID(ctx context.Context, id string) (Asset, error)
-	Upsert(context.Context, *Asset) error
+	GetIDByURN(context.Context, *Asset) (string, error)
```
StewartJingga, line 42: Do we still need this?

asset/errors.go
```diff
@@ -11,9 +12,41 @@ var (
 )
 
 type NotFoundError struct {
-	AssetID string
+	AssetID   string
+	AssetType string
```
StewartJingga, line 16: CMIIW If we don't need `AssetRepository.GetIDByURN` anymore, then we do not need `AssetType` and `AssetURN` here as well.

asset/errors.go
```diff
@@ -11,9 +12,41 @@ var (
...
+	if e.AssetURN != "" {
+		fields = append(fields, fmt.Sprintf("with asset urn \"%s\"", e.AssetURN))
+	}
+	if e.AssetType != "" {
+		fields = append(fields, fmt.Sprintf("with asset type \"%s\"", e.AssetType))
+	}
+	return fmt.Sprintf("{%s}", strings.Join(fields, " "))
+}
+
+type InvalidError struct {
```
StewartJingga, line 34: We don't need this one as well I guess along with `asset.Validate()`.

store/postgres/star_repository.go
```diff
@@ -0,0 +1,248 @@
...
+// StarRepository is a type that manages star operation to the primary database
+type StarRepository struct {
+	client *Client
+}
+
+// Create insert a new record in the stars table
+func (r *StarRepository) Create(ctx context.Context, userID string, assetID string) (string, error) {
+	var starID string
+	if userID == "" || assetID == "" {
+		return "", star.InvalidError{
```
StewartJingga, line 23: For this kind of error, I'm thinking if it would be better to use concrete error instead 🤔 . If both userID and assetID are empty, we will not get any useful information from the error.

Example
```
package star

const (
  ErrEmptyUserID = errors.New("star is not related to any user")
  ErrEmptyAssetID = errors.New("star is not related to any asset")
)
```

wdyt?

StewartJingga (changes requested):

asset/asset.go
mabdh: yes, we don't need this anymore, will revert it to the code in `main` branch

store/postgres/star_repository.go
mabdh: yes we can do that

api/handlers/user_handler.go
```diff
@@ -0,0 +1,167 @@
...
+			WriteJSONError(w, http.StatusBadRequest, err.Error())
+			return
+		}
+		if errors.As(err, new(star.NotFoundError)) {
+			WriteJSONError(w, http.StatusNotFound, err.Error())
+			return
+		}
+		internalServerError(w, h.logger, err.Error())
+		return
+	}
```
StewartJingga, line 41: Should we refactor this line to a function as it is also being used by `GetStarredAssetsWithPath` method?

mabdh: We could but I think we don't need to, unless it is being used in most places. Making this as it is so it is more readable and make the entry and exit point clearer.

StewartJingga (approved):

PR #74, opened 2022-02-02
fix: double lint check

StewartJingga:
### Problem
`lint` action is being run twice whenever there is a push on pull request.

### Root cause
It is because we set it to run with below config
```
on: [push, pull_request]
```

### Solution
Remove `pull_request` as it is already included when pushing.


StewartJingga:
@scortier

PR #81, opened 2022-02-18
feat(star): could accept email in /users/{id}/starred

mabdh:


star/star.go
```diff
@@ -21,6 +21,7 @@ type Repository interface {
 	Create(ctx context.Context, userID string, assetID string) (string, error)
 	GetStargazers(ctx context.Context, cfg Config, assetID string) ([]user.User, error)
 	GetAllAssetsByUserID(ctx context.Context, cfg Config, userID string) ([]asset.Asset, error)
+	GetAllAssetsByUserEmail(ctx context.Context, cfg Config, userID string) ([]asset.Asset, error)
```
rahmatrhd, line 24:
```go
type filter struct {
    id    string
    email string
}

type Repository interface {
    ...
	GetAllAssets(ctx context.Context, cfg Config, filter filter) ([]asset.Asset, error)
    ...
}
```

have we looked into this^ approach?

mabdh: These 2 have different sql query,
`GetAllAssetsByUserEmail` joined 3 tables
`GetAllAssetsByUserID` joined 2 tables

I am afraid it will just make it more complex

rahmatrhd (approved):

PR #86, opened 2022-03-08
feat: new lineage module with postgres

StewartJingga:
- [x] Re-enabling old lineage API for backward comp.
- [x] New `lineage` module using `lineage_graph` table.
- [x] Update `[PUT] /v1beta1/assets` to also upsert lineage using `upstreams` and `downstreams` fields.
- [x] New `[GET] /v1beta1/lineage/{urn}` to get lineage of the urn.


lineage/v2/graph.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/odpf/columbus/asset"
+
+type Graph []Edge
+
+type Edge struct {
+	// Source represents source node ID
+	Source string `json:"source"`
+
+	// Source represents source node ID
```
mabdh, line 11:
```suggestion
	// Target represents target node ID
```

store/postgres/lineage_repository.go
```diff
@@ -23,6 +24,278 @@ func NewLineageRepository(client *Client) (*LineageRepository, error) {
...
+		err = repo.removeGraph(ctx, tx, toRemoves)
+		if err != nil {
+			return fmt.Errorf("error removing graph: %w", err)
+		}
+
+		return nil
+	})
+}
+
+func (repo *LineageRepository) buildGraph(node lineage.Node, upstreams, downstreams []lineage.Node) (graph lineage.Graph) {
```
mabdh, line 70: can we move all graph processing functionality to the domain?

store/postgres/migrations/000006_create_lineage_graph_table.down.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
```
mabdh, line 1: we don't need to start transaction here since we only have 1 operation in this migration file

store/postgres/migrations/000006_create_lineage_graph_table.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
```
mabdh, line 1: same here, we don't need to start transaction here since we only have 1 operation in this migration file

api/routes.go
```diff
@@ -115,6 +124,9 @@ func RegisterRoutes(router *mux.Router, config Config) {
 	v1SubRouter := router.PathPrefix("/v1").Subrouter()
 	setupV1Beta1Router(v1SubRouter, handlerCollection)
 
+	v1Beta2SubRouter := router.PathPrefix("/v1beta2").Subrouter()
```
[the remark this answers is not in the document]
StewartJingga: `v1beta1` already has the same `/lineage/{id}` path which cannot be removed/changes at this point due to backward comp.

api/v1beta2.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/gorilla/mux"
+)
+
+func setupV1Beta2Router(router *mux.Router, handlers *Handlers) *mux.Router {
+	router.Path("/lineage/{urn}").
+		Methods(http.MethodGet).
+		HandlerFunc(handlers.LineageV2.GetGraph)
+
+	return router
```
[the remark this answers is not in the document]
StewartJingga: oh sure, was copying from `setupV1Beta1Router`, should I remove that as well?

api/handlers/lineage_handler.go
```diff
@@ -31,6 +35,30 @@ func NewLineageHandler(logger log.Logger, provider LineageProvider) *LineageHand
 	return handler
 }
 
+func (handler *LineageHandler) ListLineage(w http.ResponseWriter, r *http.Request) {
+	graph, err := handler.lineageProvider.Graph()
+	if err != nil {
+		handler.logger.Error("error requesting graph", "error", err)
+		status := http.StatusInternalServerError
+		WriteJSONError(w, status, http.StatusText(status))
```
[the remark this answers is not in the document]
StewartJingga: oh actually this line was from a reverse commit to enable back old lineage API.
will change it to use latest approach.

store/postgres/lineage_repository.go
StewartJingga: we could but this one is specific function of the repository implementation (postgres) not a generic one.

if we are making this generic, then it would also makes sense to change the `lineage.Repository` contract so that it receives `lineage.Graph` when upserting as opposed to `[]lineage.Node`.

in summary, it depends on how we are creating our `lineage.Repository` contract, but if we are using the current contract, moving it to domain can give wrong idea.

api/routes.go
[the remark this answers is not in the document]
mabdh: Are we already using `v1beta1` in old datlantis?

We have `v1` endpoint that I think is being used by datlantis and is related with old behaviour.

I am thinking maybe better to leave all old behaviour in `v1`. If we have a plan to deprecate that in the future and use `v1beta1` for the new behaviour.

I am afraid adding a new version `v1beta2` for lineage only would make things more confusing. wdyt?

api/v1beta2.go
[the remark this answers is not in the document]
mabdh: yes please stew 🙏🏼

store/postgres/lineage_repository.go
mabdh: oh yes, that is correct, let's keep it like this then

api/routes.go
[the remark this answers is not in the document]
StewartJingga: unfortunately old datlantis is already hitting `/v1beta1`, putting the new lineage on `/v1beta1` will introduce breaking changes and more effort on datlantis side.

so the options are whether we have (`v1` and `v1beta1`) or (`v1beta1` and `v1beta2`). Both are fine to me but the former will introduce breaking changes and efforts on datlantis. wdyt @mabdh ?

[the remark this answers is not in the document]
mabdh: I see, it is okay then I think to have `v1beta1` and `v1beta2`
any thought on this @ravisuhag ?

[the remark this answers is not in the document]
mabdh: Let's make it `v1beta1` stew? @StewartJingga

api/v1beta1.go
```diff
@@ -19,16 +19,21 @@ func setupV1Beta1Router(router *mux.Router, handlers *Handlers) *mux.Router {
 		Methods(http.MethodGet).
 		HandlerFunc(handlers.Search.Suggest)
 
-	router.PathPrefix("/lineage/{id}").
+	router.PathPrefix("/lineage/{urn}").
 		Methods(http.MethodGet).
-		HandlerFunc(handlers.Lineage.GetLineage)
+		HandlerFunc(handlers.LineageV2.GetGraph)
```
mabdh, line 24: @StewartJingga if we are going to remove a whole functionality of old Lineage, wouldn't it be better to remove the package as well? So we don't have `lineage/v1` and `lineage/v2`. Package `lineage` will only have the latest version one.

ravisuhag: +1

StewartJingga: yeah missed that, thanks for catching that, removed!

mabdh:
LGTM

mabdh:
LGTM

StewartJingga:
> LGTM
> 
> but first can we create an issue first and link to this PR? @StewartJingga

Done!

PR #87, opened 2022-03-09
feat: discussions feature

mabdh:
- [x] List all discussions with filter API
- [x] Create a new discussion API
- [x] Patch a discussion API
- [x] Get a discussion API


discussion/filter.go
```diff
@@ -0,0 +1,64 @@
...
+	SortBy        string `json:"sort" validate:"omitempty,oneof=created_at updated_at"`
+	SortDirection string `json:"direction" validate:"omitempty,oneof=asc desc"`
+	Size          int
+	Offset        int
+}
+
+// Validate will check whether fields in the filter fulfills the constraint
+func (f *Filter) Validate() error {
+
+	if f.Size < 0 {
```
StewartJingga, line 41: is there a reason on why are we not using validate annotation like some of the fields?

discussion/filter.go
```diff
@@ -0,0 +1,64 @@
...
+		if name == "-" {
+			return ""
+		}
+		return name
+	})
+}
+
+type Filter struct {
+	Type          string `json:"type" validate:"omitempty,oneof=openended issues qanda all"`
+	State         string
```
StewartJingga, line 27: like `Type` should we also validate `State`?

discussion/state.go
```diff
@@ -0,0 +1,54 @@
+package discussion
+
+type State int
```
StewartJingga, line 3: if we are going with `State` as a string instead of int, will that be simpler?
Or is there other concerns?

discussion/types.go
```diff
@@ -0,0 +1,60 @@
+package discussion
+
+type Type int
```
StewartJingga, line 3: Similar with `State`, will changing it to `string` makes it simpler?

store/postgres/asset_repository.go
```diff
@@ -47,7 +47,6 @@ func (r *AssetRepository) GetAll(ctx context.Context, config asset.Config) (asse
 		return
 	}
 
-	assets = []asset.Asset{}
```
StewartJingga, line 50: oh actually this is to prevent `GetAll` to return `null` to client as opposed to empty array.

store/postgres/discussion_model.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/lib/pq"
+	"github.com/odpf/columbus/discussion"
+)
+
+type DiscussionModel struct {
+	ID        string         `db:"id"`
+	Title     string         `db:"title"`
+	Body      string         `db:"body"`
+	Type      int            `db:"type"`
+	State     int            `db:"state"`
```
StewartJingga, line 15: I do know storing this as number will be much cheaper format in DB but aside from that is there any other concerns?
Storing these values as string may take more space but should be simpler to implement all around the code and hopefully better development/debugging experience.

From my exp, dbsize rarely matters in this case unless im missing something.

store/postgres/migrations/000006_create_discussions_table.up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE discussions (
+  id serial PRIMARY KEY,
+  title text NOT NULL,
+  body text NOT NULL,
+  state integer NOT NULL DEFAULT 0,
+  type integer NOT NULL DEFAULT 0,
+  owner uuid NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
StewartJingga, line 7: hmm im thinking to restrict it instead of cascade, missing discussion can be confusing later on.
just a thought, wdyt?

discussion/filter.go
mabdh: Just want to make the error specific and not returning error from Validator to the user.

discussion/filter.go
mabdh: Oh yes thanks for pointing out, previously state would depend on the type, but now it is possible to check it since all type will have the same state names.

mabdh: Oh yes thanks for pointing out, previously state would depend on the type, but now it is possible to check it since all type will have the same state names.

discussion/state.go
mabdh: The reason why integer is being used to store the state is to decouple it from the storage and api. State in the API needs to be readable. This is to give more flexibility if we want to name the state differently for each type.

store/postgres/asset_repository.go
mabdh: I see, so I think we can make it simpler by not using `named parameter` in the return to make the flow clearer

store/postgres/discussion_model.go
mabdh: It is actually not about the db size but more into having no semantic stored in the DB.

It is actually possible to use it as string, let me check.

store/postgres/migrations/000006_create_discussions_table.up.sql
mabdh: if we restrict it, we can't remove user id in the users table. If that's the reason, we can remove the foreign key of this. So, discussion is immutable.

StewartJingga: hmm yes we won't be to remove a user if he/she has created a discussion. I'm still comparing the downside is better if we cascadingly delete a discussion which could have evolved and store an important information.

or we can use `set null`? wdyt?

mabdh: Yes discussion should not be deleted.

Hmm I don't think `null` would help. I think ownership information should always be there regardless user exists or not in user table. It is okay I guess to not have FK. The worst case would be we have empty `user.User` when doing left join the discussion with user table.

store/postgres/asset_repository.go
StewartJingga: yeah sure, if it is bringing confusion the it is definitely needs to be refactored.

store/postgres/discussion_model.go
StewartJingga: Hmm ok, personally I would go for storing enum as string first until it is not feasible anymore (which rarely happens) because it is easier and quicker to implement.

But since we already have all the codes written then it would be counter productive to revert it again.
Also CMIIW, we need to make sure the order of the enums are not changed or else we could get corrupted data which is hard to fix. (someone's creating a discussion with application having wrong enum order).

mabdh: Sure I am updating it to string, it is not that complex though.

StewartJingga: Oh it is fine, since you already implemented those. We can keep storing enum as int. @mabdh

mabdh: @StewartJingga already updated it to string 😄

StewartJingga: oh okay 😅 , sorry for the trouble 🙏 .

StewartJingga (approved):
LGTM

StewartJingga (approved):
LGTM

PR #91, opened 2022-03-11
feat: discussions comments and user/discussions api

mabdh:


StewartJingga (commented):
Should we put comment in discussion package as it fully depends on discussion. We can also refer is as discussion.Comment.

Wdyt?

mabdh:
@StewartJingga yes we can do that too, no problem with that

mabdh:
@StewartJingga but after thinking again, I think better to keep it in its own package. Even if it is fully dependent, would that be a problem?

StewartJingga:
> @StewartJingga I think better to keep it in its own package. Even if it is fully dependent, would that be a problem?
> 
> My concern is, we need to have `discussion.DiscussionRepository` and `discussion.CommentRepository` instead of `discussion.Repository` alone.

We can still keep `discussion.Repository` since the package name is discussion. Also we might not even need `discussion.CommentRepository`, we could have `AddComment()` method on `discussion.Repository`.

My concerns are:
1. Having comment as its own package could give a wrong impression that it is a standalone feature while it is not.
2. Having it under a discussion makes sense because they are the same domain and the hard dependency proves that.
3. Lastly (this is personal preference, can be ignored), discussion and comment are one feature and it feels good to know that we can port discussion feature to other codebase/module with just moving one folder (package/module).

mabdh:
> Having comment as its own package could give a wrong impression that it is a standalone feature while it is not.
It wouldn't really matter whether it is standalone feature or not right? We better keep the domain as granular as possible. I don't think combining comment inside discussion would matter that much.

> Having it under a discussion makes sense because they are the same domain and the hard dependency proves that.
Yes it does make sense but again I don't think having hard dependency would really matter here

> Lastly (this is personal preference, can be ignored), discussion and comment are one feature and it feels good to know that we can port discussion feature to other codebase/module with just moving one folder (package/module).
I agree on this. But it is also would make it easier to copy comment package if we need comments feature only in other use cases.

Anyway, let me try updating it according to your sugggestion.

mabdh:
@StewartJingga comments moved to discussion package, please review it again stew :pray:

store/postgres/migrations/000008_create_comments_table.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE comments (
+  id serial PRIMARY KEY,
+  discussion_id serial NOT NULL REFERENCES discussions(id) ON DELETE CASCADE,
+  body text NOT NULL,
+  owner uuid NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
StewartJingga, line 5: will it be better to name it `user_id` instead in the DB?

StewartJingga (changes requested):

discussion/errors.go
```diff
@@ -13,9 +13,33 @@ var (
...
 	DiscussionID string
 }
 
 func (e NotFoundError) Error() string {
-	return fmt.Sprintf("could not find discussion with id: %s", e.DiscussionID)
+	fields := []string{"could not find"}
+	if e.DiscussionID != "" {
+		fields = append(fields, fmt.Sprintf(" discussion with id \"%s\"", e.DiscussionID))
+	}
+	if e.CommentID != "" {
```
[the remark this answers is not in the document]
mabdh: Oh yes, I need to fix the error.

For now there is no specific cases where we need to have `CommentNotFoundError`, I think better to make it `NotFoundError` only for now.

store/postgres/migrations/000008_create_comments_table.up.sql
mabdh: do you mean like this?
```suggestion
  user_id uuid NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
IMO `user_id` is not self-explanatory, it could be anything. I am using owner because we are consistently using `owner` for all resources.

filter/validator.go
```diff
@@ -0,0 +1,73 @@
+package filter
```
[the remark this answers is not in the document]
mabdh: I was thinking that way as well. But we also have `validator` package in `tag`. We can use `validator` for this but maybe in the future we can think about how we centralize it.

store/postgres/migrations/000008_create_comments_table.up.sql
StewartJingga: no major issue actually, just used to convention having FK as `{singular_table_name}_id`.
we can keep owner, should be fine.

swagger.yaml
```diff
@@ -264,6 +264,412 @@ paths:
...
+                description: id of the newly created discussion
+        400:
+          description: validation error
+          schema:
+            $ref: "#/definitions/ValidationError"
+        500:
+          description: "internal server error"
+          schema:
+            $ref: "#/definitions/InternalServerError"
+  "/v1beta1/discussions/{id}":
```
StewartJingga, line 389: this is duplicate key, cannot run swagger due to this.

swagger.yaml
```diff
@@ -264,6 +264,412 @@ paths:
...
+          type: integer
+          required: true
+        - in: body
+          name: ""
+          required: true
+          schema:
+            allOf:
+              - $ref: "#/definitions/Comment"
+              - required:
+                  - body
```
StewartJingga, line 537: Since only `body` field is required here let's just change it like this to prevent misleading user having to prepare all other fields.

```suggestion
            type: object
            properties:
              body:
                type: string
                example: "This is my first comment"
                required: true
```

StewartJingga (changes requested):
Small issue on swagger file

swagger.yaml
mabdh: Thanks stew, fixing this @StewartJingga

StewartJingga (approved):
LGTM, thanks for also removing old lineage api from swagger

PR #97, opened 2022-03-18
feat: add grpc-gateway

mabdh:


grpc_interceptor/user.go
```diff
@@ -0,0 +1,41 @@
+package grpc_interceptor
```
StewartJingga, line 1: should we move this package to be in the `api` folder since it is being used there and it has  grpc request/response context?

third_party/OpenAPI/compass.swagger.json
```diff
@@ -0,0 +1,717 @@
+{
```
StewartJingga, line 1: is this auto-generated?

mabdh: Yes stew

grpc_interceptor/user.go
mabdh: Yes that could be done

third_party/OpenAPI/compass.swagger.json
StewartJingga: awesome!

grpc_interceptor/user.go
mabdh: @StewartJingga after checking again through the code, this is being used in `cmd/server.go` but I think this still could be moved to the `api` package

StewartJingga:
Need help updating `config.yaml.example` as well for the new config `GRPC_SERVER_PORT`. 🙏

mabdh:
@StewartJingga it is for temporary stew

PR #98, opened 2022-03-18
feat: new upsert patch API

StewartJingga:
- [x] new Asset.Patch method to merge assets
- [x] create new api [PATCH] /v1beta1/assets
- [x] update swagger

### DIFF
--- a/api/handlers/asset_handler.go
+++ b/api/handlers/asset_handler.go
@@ -134,7 +134,67 @@ func (h *AssetHandler) Upsert(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := h.saveLineage(r.Context(), payload); err != nil {
+	if err := h.saveLineage(r.Context(), ast, payload.Upstreams, payload.Downstreams); err != nil {
+		internalServerError(w, h.logger, err.Error())
+		return
+	}
+
+	writeJSON(w, http.StatusOK, map[string]interface{}{
+		"id": ast.ID,
+	})
+}
+
+func (h *AssetHandler) UpsertPatch(w http.ResponseWriter, r *http.Request) {
+	userID := user.FromContext(r.Context())
+	if userID == "" {
+		h.logger.Warn(errMissingUserInfo.Error())
+		WriteJSONError(w, http.StatusBadRequest, errMissingUserInfo.Error())
+		return
+	}
+
+	var payload patchAssetPayload
+	err := json.NewDecoder(r.Body).Decode(&payload)
+	if err != nil {
+		WriteJSONError(w, http.StatusBadRequest, bodyParserErrorMsg(err))
+		return
+	}
+
+	urn, typ, service, err := h.validatePatchPayload(payload.Asset)
+	if err != nil {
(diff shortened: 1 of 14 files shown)


api/handlers/asset_handler.go
```diff
@@ -134,7 +134,67 @@ func (h *AssetHandler) Upsert(w http.ResponseWriter, r *http.Request) {
...
+		internalServerError(w, h.logger, err.Error())
+		return
+	}
+
+	writeJSON(w, http.StatusOK, map[string]interface{}{
+		"id": ast.ID,
+	})
+}
+
+func (h *AssetHandler) UpsertPatch(w http.ResponseWriter, r *http.Request) {
```
mabdh, line 147: Should we just say patch? to avoid confusion

asset/errors.go
```diff
@@ -13,10 +13,17 @@ var (
...
+	Service string
 }
 
 func (err NotFoundError) Error() string {
-	return fmt.Sprintf("no such record: %q", err.AssetID)
+	if err.AssetID != "" {
+		return fmt.Sprintf("no such record: %q", err.AssetID)
+	} else {
+		return fmt.Sprintf("could not find asset with urn = %s, type = %s, service = %s", err.URN, err.Type, err.Service)
+	}
```
mabdh, line 26:
```suggestion
	if err.AssetID != "" {
		return fmt.Sprintf("no such record: %q", err.AssetID)
	} 
        return fmt.Sprintf("could not find asset with urn = %s, type = %s, service = %s", err.URN, err.Type, err.Service)
```

Can also make it more compact

api/v1beta1.go
```diff
@@ -42,6 +42,10 @@ func setupV1Beta1AssetRoutes(baseURL string, router *mux.Router, ah *handlers.As
 		Methods(http.MethodPut, http.MethodHead).
 		HandlerFunc(ah.Upsert)
 
+	router.Path(baseURL).
+		Methods(http.MethodPatch, http.MethodHead).
```
mabdh, line 46: any specific reason we use `HEAD` here?

api/handlers/asset_handler.go
StewartJingga: was thinking about that, but this also allows creation.

api/v1beta1.go
StewartJingga: was being brought from legacy code and copypasted over from other routes, wasn't really sure the first reason we had that.

🤔  Should we remove it from all routes?

mabdh: @StewartJingga afaik `HEAD` is a short version of `GET`, usually it is being used for resource/system health checking (figuring out whether the docs exist or not). For example, in Elasticsearch afaik before the client connect to the cluster, it sends HEAD first.

Even if we want to keep it, would be more make sense if we keep it in `GET` only. I think we need to remove it from the others PUT/PATCH.

Since we are not really have a behaviour of `HEAD` in our API, removing all of them would be better.

mabdh:
LGTM

PR #100, opened 2022-03-21
feat: add filtering in assets api

scortier:
### Added filtering by certain fields in assets api.
`Querystring:  "?types=table&services=bigquery&size=30&offset=50&sort=created_at&direction=desc&filter[data.dataset]=booking&filter[data.project]=p-godata-id&q=internal&q_fields=name,urn,description,services",`

- filter by types : `types=topic,table`
- filter by services: `services=kafka,postgres`
- filter by field in asset.data: `filter[data.dataset]=booking&filter[data.project]=p-godata-id"`
- querying by field ( includes by name , urn): `q=internal&q_fields=name,urn,description,services`
- sort by certain fields : `sort=created_at`
- sorting direction (asc / desc) : `direction=desc`


store/postgres/asset_repository.go
```diff
@@ -25,27 +25,27 @@ type AssetRepository struct {
...
 	}
-	ams := []*AssetModel{}
+
+	var ams []*AssetModel
 	err = r.client.db.SelectContext(ctx, &ams, query, args...)
 	if err != nil {
 		return nil, fmt.Errorf("error getting asset list: %w", err)
 	}
 
-	assets := []asset.Asset{}
```
StewartJingga, line 48: we need this line `assets := []asset.Asset{}` to make GetAll to return empty slice instead of null if there are 0 records found.

store/postgres/asset_repository.go
```diff
@@ -649,6 +634,46 @@ func (r *AssetRepository) getAssetVersionSQL() sq.SelectBuilder {
...
+}
+
+func (r *AssetRepository) buildOrderQuery(builder sq.SelectBuilder, cfg asset.Config) sq.SelectBuilder {
+	if cfg.SortBy != "" {
+		orderDirection := "ASC"
+		if cfg.SortDirection != "" {
+			orderDirection = strings.ToUpper(cfg.SortDirection)
+		}
+		return builder.OrderBy(cfg.SortBy + " " + orderDirection)
+	}
```
StewartJingga, line 672: Let's try to return early to avoid nested indentation.

```suggestion
        if cfg.SortBy == "" {
                return builder
        }
        
        orderDirection := "ASC"
	if cfg.SortDirection != "" {
	        orderDirection = cfg.SortDirection
	}
			
	return builder.OrderBy(cfg.SortBy + " " + orderDirection)
```

store/postgres/asset_repository_test.go
```diff
@@ -18,6 +20,8 @@ import (
 
 var defaultAssetUpdaterUserID = uuid.NewString()
 
+const defaultGetAssetMaxSize = 3
```
StewartJingga, line 23: do we need this variable here? it can be moved to the test using it only.

store/postgres/asset_repository_test.go
```diff
@@ -91,46 +95,54 @@ func (r *AssetRepositoryTestSuite) TearDownSuite() {
...
-		for i := 0; i < defaultGetMaxSize; i++ {
+		r.Require().Len(results, defaultGetAssetMaxSize)
+		for i := 0; i < defaultGetAssetMaxSize; i++ {
 			r.assertAsset(&assets[i], &results[i])
 		}
 	})
 
 	r.Run("should override default size using GetConfig.Size", func() {
-		size := 8
+		size := 3
```
StewartJingga, line 145: this test won't confirm anything, defaultGetAssetMaxSize is `3` and size is `3`, and the total assets in fixture is also `3`.

api/httpapi/handlers/asset_handler.go
```diff
@@ -61,9 +69,8 @@ func (h *AssetHandler) GetAll(w http.ResponseWriter, r *http.Request) {
 	withTotal, ok := r.URL.Query()["with_total"]
 	if ok && len(withTotal) > 0 && withTotal[0] != "false" && withTotal[0] != "0" {
 		total, err := h.assetRepo.GetCount(r.Context(), asset.Config{
```
StewartJingga, line 71: `GetCount` needs to have the same config value as `GetAll`, so we won't get the wrong total.

api/httpapi/handlers/asset_handler_test.go
```diff
@@ -676,37 +676,60 @@ func TestAssetHandlerGet(t *testing.T) {
 			Description:  `should return http 500 if fetching fails`,
 			ExpectStatus: http.StatusInternalServerError,
 			Setup: func(ctx context.Context, ar *mocks.AssetRepository) {
-				ar.On("GetAll", ctx, asset.Config{}).Return([]asset.Asset{}, errors.New("unknown error"))
+				ar.On("GetAll", ctx, asset.Config{
+					Data: make(map[string]string),
```
StewartJingga, line 680: instead of doing this, is it possible to change the implementation to not instantiate a map if not needed?

api/httpapi/handlers/asset_handler_test.go
```diff
@@ -676,37 +676,60 @@ func TestAssetHandlerGet(t *testing.T) {
...
+				}).Return([]asset.Asset{}, nil, nil)
 				ar.On("GetCount", ctx, asset.Config{}).Return(0, errors.New("unknown error"))
 			},
 		},
 		{
 			Description:  `should parse querystring to get config`,
-			Querystring:  "?text=asd&type=table&service=bigquery&size=30&offset=50",
+			Querystring:  "?types=table&services=bigquery&size=30&offset=50&sort=created_at&direction=desc&config.data[dataset]=booking&config.data[project]=p-godata-id",
```
StewartJingga, line 697: need help adding more test cases for:
1. `q` and `q_fields` querystring
2. `data` querystring

You can create new test cases if you want instead of piling them all here.

StewartJingga (changes requested):
Also we need to update `AssetRepository.GetCount` to also read and process the same filter config as `GetAll`. Or else we will get wrong total value.

swagger.yaml
```diff
@@ -1634,14 +1666,15 @@ definitions:
         type: string
         example: "sample-name"
       service:
-        type: string
-        example: "bigquery"
+        type: array
+        example: "bigquery,postgres"
```
mabdh, line 1670: @scortier are we also updating the asset type to be multiple types here?

swagger.yaml
```diff
@@ -1634,14 +1666,15 @@ definitions:
...
       description:
         type: string
         example: "sample description"
       data:
         type: object
         description: "dynamic object"
+        example: "config.data[entity]=gojek"
```
mabdh, line 1677: @scortier I don't think this query like structure belongs to this place. The `config.data[entity]=gojek` should only come in the query param right?

mabdh:
@scortier 
```
filter by field in asset.data: config.data[dataset]=booking&config.data[project]=p-godata-id"
```
If it is for filtering, why not using `filter` prefix instead?

---
The nested query params is in the wrong format.
`config.data[dataset]=booking&config.data[project]=p-godata-id"` ❌ 
`config[data.dataset]=booking&config[data.project]=p-godata-id"` ✅ 
`filter[data.dataset]=booking&filter[data.project]=p-godata-id"` ✅  ✅ 

cc @StewartJingga

swagger.yaml
scortier: Oh, the base asset is not changed, thanks for pointing out, will revert the changes

scortier:
@mabdh @StewartJingga,  the proto file needs to be updated to pass correct config in `GetAllAsset` method in `v1beta1/asset`, due to this reason the test is failing, shall we update the proto file and fix this test or move forward with failing test, as fixing the proto file is in scope of [this issue](https://github.com/odpf/columbus/issues/105). 
WDYT ?

StewartJingga:
> @scortier
> 
> ```
> filter by field in asset.data: config.data[dataset]=booking&config.data[project]=p-godata-id"
> ```
> 
> If it is for filtering, why not using `filter` prefix instead?
> 
> The nested query params is in the wrong format. `config.data[dataset]=booking&config.data[project]=p-godata-id"` ❌ `config[data.dataset]=booking&config[data.project]=p-godata-id"` ✅ `filter[data.dataset]=booking&filter[data.project]=p-godata-id"` ✅ ✅
> 
> cc @StewartJingga

Why there is `config` there? We can directly use `data[]` instead. @scortier

StewartJingga:
> @mabdh @StewartJingga, the proto file needs to be updated to pass correct config in `GetAllAsset` method in `v1beta1/asset`, due to this reason the test is failing, shall we update the proto file and fix this test or move forward with failing test, as fixing the proto file is in scope of [this issue](https://github.com/odpf/columbus/issues/105). WDYT ?

We can just fix the proto in this PR instead, we dont want breaking test on main branch. @scortier

scortier:
> > @scortier
> > ```
> > filter by field in asset.data: config.data[dataset]=booking&config.data[project]=p-godata-id"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If it is for filtering, why not using `filter` prefix instead?
> > The nested query params is in the wrong format. `config.data[dataset]=booking&config.data[project]=p-godata-id"` ❌ `config[data.dataset]=booking&config[data.project]=p-godata-id"` ✅ `filter[data.dataset]=booking&filter[data.project]=p-godata-id"` ✅ ✅
> > cc @StewartJingga
> 
> Why there is `config` there? We can directly use `data[]` instead. @scortier

Recently, updated the query param to this format : `filter[data.dataset]=booking&filter[data.project]=p-godata-id"`, should we update it to `data[dataset]=booking&data[project]=p-godata-id"` ?, but we discuss earlier the format should be of type `filter[{fields}]`.  @mabdh @StewartJingga

mabdh:
For the proto, in that case I think this PR could also covers this issue https://github.com/odpf/columbus/issues/105 @scortier @StewartJingga

StewartJingga (approved):
LGTM

PR #107, opened 2022-04-06
feat: add nested query params in Get all assets API

scortier:
### The columbus should support nesting query in Data and Query Filtering :
- For Data Filter : `data[entity.properties.landscape]=internal`
- For Query Filter : `q=internal&q_fields=data.entity.properties.landscape`


store/postgres/asset_repository.go
```diff
@@ -634,6 +634,7 @@ func (r *AssetRepository) getAssetVersionSQL() sq.SelectBuilder {
 		LeftJoin("users u ON a.updated_by = u.id")
 }
 
+// buildFilterQuery retrieves the sql query based on applied filter in the queryString
 func (r *AssetRepository) buildFilterQuery(builder sq.SelectBuilder, cfg asset.Config) sq.SelectBuilder {
```
mabdh, line 638: this looks like a complex query builder. 
can we have a test for this?

mabdh (changes requested):

store/postgres/asset_repository.go
scortier: I think every single query builder in `AssetRepository.buildFilterQuery` method is tested in `AssetRepositoryTestSuite.TestGetAll` method. So does it will create any difference on testing it again.
WDYT ??

mabdh: @scortier true but since it has a complex logic, better to be tested again alone. `Given the config, what psql query do we expect`

scortier: okay, sure will add it 👍🏻

store/postgres/asset_repository_test.go
```diff
@@ -127,6 +130,104 @@ func (r *AssetRepositoryTestSuite) insertRecord() (assets []asset.Asset) {
...
+			Data: map[string]string{
+				"landscape.properties.project-id": "columbus_001",
+				"country":                         "vn",
+			},
+		})
+		query, _, err := results.ToSql()
+		r.Require().NoError(err)
+		query, err = sq.Dollar.ReplacePlaceholders(query)
+		expectedQuery := strings.Split(query, "WHERE ")
+		actualQuery := `data->'landscape'->'properties'->>'project-id' = 'columbus_001' AND data->>'country' = 'vn'`
```
mabdh, line 226: wouldn't it be like this?
```suggestion
		expectedQuery := `data->'landscape'->'properties'->>'project-id' = 'columbus_001' AND data->>'country' = 'vn'`
```

store/postgres/asset_repository_test.go
```diff
@@ -127,6 +130,104 @@ func (r *AssetRepositoryTestSuite) insertRecord() (assets []asset.Asset) {
 	return assets
 }
 
+func (r *AssetRepositoryTestSuite) TestBuildFilterQuery() {
+	r.builder = sq.Select(`a.test as test`)
+
+	r.Run("should not return all assets limited by default size", func() {
```
mabdh, line 136: actually you can put these tests into a [table driven test](https://dev.to/boncheff/table-driven-unit-tests-in-go-407b) to simplify it. You can check on of the tests in columbus repo.

mabdh (changes requested):

mabdh:
@scortier code looks good, is it tested in your local?

StewartJingga (approved):

rsbh (approved):

mabdh (approved):

PR #110, opened 2022-04-12
feat(user): add uuid to user information

mabdh:


store/postgres/migrations/000002_create_users_table.up.sql
```diff
@@ -1,7 +1,11 @@
 CREATE TABLE users (
     id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
-    email text UNIQUE NOT NULL,
-    provider text NOT NULL,
+    uuid text,
```
StewartJingga, line 3: if we are changing here, then it means we need to clear/reset the DB that has undergone further migration right?

mabdh: yes and for now I guess it is fine since our users feature and feature that use DB usage is still not final.

StewartJingga: ack, for now I believe it is still okay.

We may need to decide for next time whether we should consider breaking main branch migration or not.

StewartJingga (approved):

PR #122, opened 2022-05-11
feat(tags): replace tags identifier from record urn and type to asset id

mabdh:
- replace tags identifier from record urn and type to asset id 
- update docs


StewartJingga (approved):
LGTM

PR #126, opened 2022-05-13
feat: add cli commands for all compass APIs

GrayFlash:


cmd/assets.go
```diff
@@ -13,17 +13,17 @@ import (
 
 func assetsCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "asset",
```
ravisuhag: @GrayFlash  it should be singular only. we can add plural as alias though.

cmd/assets.go
```diff
@@ -13,17 +13,17 @@ import (
...
+		Use:     "assets",
 		Aliases: []string{"assets"},
 		Short:   "Manage assets",
 		Annotations: map[string]string{
 			"group:core": "true",
 		},
 		Example: heredoc.Doc(`
-			$ compass asset list
-			$ compass asset get
```
ravisuhag: We can use `view` as a command name here. and edit for post.

cmd/assets.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+			defer cancel()
+
+			ctx := setCtxHeader(cmd.Context(), header)
+			res, err := client.GetAllAssets(ctx, &compassv1beta1.GetAllAssetsRequest{})
+			if err != nil {
+				return err
+			}
+
+			fmt.Println(cs.Bluef(prettyPrint(res.GetData())))
```
StewartJingga, line 65: instead of by default printing it as json, should we print it in a table format

example:
```
URN           NAME             TYPE        SERVICE
some-urn      some-name        table       bigquery
```
we still also provide an alternative to output as json tho, but by default we can start with table first.

wdyt? @ravisuhag

StewartJingga (changes requested):

cmd/config.go
```diff
@@ -11,8 +11,10 @@ import (
 
 type Config struct {
 	// Server Config
-	ServerHost string `mapstructure:"SERVER_HOST" default:"0.0.0.0"`
-	ServerPort string `mapstructure:"SERVER_PORT" default:"8080"`
+	ServerHost    string `mapstructure:"SERVER_HOST" default:"0.0.0.0"`
+	ServerPort    string `mapstructure:"SERVER_PORT" default:"8080"`
+	ServerBaseUrl string `mapstructure:"SERVER_BASE_URL" default:"localhost:8080"`
+	AuthHeader    string `mapstructure:"AUTH_HEADER" default:"Compass-User-UUID:odpf@email.com"`
```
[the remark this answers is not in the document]
GrayFlash: Yes, this seems better

cmd/assets.go
GrayFlash: I can agree that the table is usually more presentable, but here since we are developing a cli that supports us with making requests to compass, I think printing it in `json` should be fine, since it is usually how most of the responses. Have tried to keep the output presentable on our part. 

Also, I think if we represent this response in the table it will be a bit difficult to maintain all the key terms in a column. for e.g. for the below response:

```
{
		"id": "dffbe72e-96e6-4c1e-8752-6bee19c2fa11",
		"urn": "myjob::tableb",
		"type": "job",
		"service": "myjob",
		"name": "ironman_extractor",
		"version": "0.1",
		"updated_by": {
			"uuid": "stewart.jingga@gojek.com"
		},
		"created_at": {
			"seconds": 1652932353,
			"nanos": 731728000
		},
		"updated_at": {
			"seconds": 1652932353,
			"nanos": 731728000
		}
	}
```

I feel that things like updated_by, and assets array in discussion will tough to present in table. Also If we add more fields in response in an asset later, we will need to update cmd/assets.go as well.

cmd/assets.go
GrayFlash: `view` seems better than get. But should we use something like `upsert` or `add` instead for post, edit seems like making changes to something existing. wdyt @ravisuhag ??

GrayFlash:
Currently we have added support for the following set of APIs:
- [ ] assets (list, view, delete, post)
- [ ] discussion (list, view, post)
- [ ] search <text> (with flags as filters, query, rankBy and size)
- [ ] lineage <urn>

And further APIs that we are planning to support over here @StewartJingga @ravisuhag ??

cmd/assets.go
StewartJingga: We can do it like how k8s does. If we are outputing json format as default, imho it will be too overwhelming and close to not being usable unless it is being piped to a file. Not to mention we have lots of record.

With table, we only show relevant information and as a quick look on the assets. We will still allow output to be in json format with a flag e.g. `-o json`.

cmd/assets.go
GrayFlash: Sorry, just noticed that `upsert` has been deprecated, edit make sense

cmd/assets.go
GrayFlash: Have added this, take a look once.

Screenshot 2022-05-23 at 8 18 31 AM

.gitignore
```diff
@@ -23,5 +23,10 @@ vendor/
 
 # Project specific ignore
 .env
+asset-body.yaml
+asset-body.json
```
mabdh, line 27: why we need this here?

cli/discussions.go
```diff
@@ -0,0 +1,185 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&json, "out", "o", "table", "flag to control output viewing, for json `-o json`")
+
+	return cmd
+}
+
+func viewdiscussionByIDCommand() *cobra.Command {
```
mabdh, line 89:
```suggestion
func viewDiscussionByIDCommand() *cobra.Command {
```

cli/discussions.go
```diff
@@ -0,0 +1,185 @@
...
+
+			fmt.Println(cs.Bluef(prettyPrint(res.GetData())))
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func postdiscussionCommand() *cobra.Command {
```
mabdh, line 128:
```suggestion
func postDiscussionCommand() *cobra.Command {
```

compass.yaml.example
```diff
@@ -17,14 +17,16 @@ db:
...
-    user: postgres
-    password: ~
+    user: compass
+    password: compass_password
     sslmode: disable
 
 service:
     host: localhost
     port: 8080
+    baseurl: localhost:8080
```
mabdh, line 27: wouldn't this make it confusing to have different config of server address for client and server?

GrayFlash: Yes, moved it to the client when pairing with mas stew.

.gitignore
GrayFlash: Was there for me to not push asset-body and discussion-body file to github.. Have removed it now.

StewartJingga (approved):

PR #132, opened 2022-05-19
fix: asset filter query test

scortier:


store/postgres/asset_repository_test.go
```diff
@@ -169,25 +169,25 @@ func (r *AssetRepositoryTestSuite) TestBuildFilterQuery() {
 			},
 			expectedQuery: `(data->'landscape'->'properties'->>'project-id' ILIKE $1 OR description ILIKE $2)`,
 		},
+		// NOTE: The data entries contain in the map does not produce any specific order for query, generated query
```
StewartJingga, line 172: The comment can focus more on the limitation on golang's map, for example:
```
// NOTE: Cannot have more than one key in because golang's map does not guarantee order thus producing inconsistent test.
```
And we can put the comment on the Data filter itself, to emphasize it.
wdyt?

StewartJingga (changes requested):

StewartJingga (approved):

PR #136, opened 2022-05-25
feat(elasticsearch): make service as an index

mabdh:


internal/store/elasticsearch/discovery_repository.go
```diff
@@ -82,9 +83,9 @@ func (repo *DiscoveryRepository) Delete(ctx context.Context, assetID string) err
 }
 
 func (repo *DiscoveryRepository) GetTypes(ctx context.Context) (map[asset.Type]int, error) {
```
StewartJingga, line 85: I wonder if we still need this 🤔.
Do you think it would be better if we get types count from asset repository (postgres) instead?

mabdh: Yes I also thought that way. I was doing this only to make the changes as simple as possible to avoid some regression errors.

But I think we could move to postgres count for this. Will do the changes, thanks stew.

StewartJingga: @mabdh it could be another story/issue actually. As far as this PR goes, it looks good to me. wdyt? @mabdh

mabdh: @StewartJingga fair enough, let's create a new issue for that stew

StewartJingga (approved):

PR #144, opened 2022-06-02
fix: misleading error when sending different user id for the same email

scortier:


internal/store/postgres/user_repository.go
```diff
@@ -27,8 +27,7 @@ func (r *UserRepository) UpsertByEmail(ctx context.Context, ud *user.User) (stri
 	um := newUserModel(ud)
 
 	if err := r.client.db.QueryRowxContext(ctx, `
-				INSERT INTO users (uuid, email, provider) VALUES ($1, $2, $3) ON CONFLICT (email)
```
mabdh, line 30: The logic here supposed to be not changed. 

Please read the explanation about how compass handles user [here](https://github.com/odpf/compass/blob/main/docs/concepts/user.md)

mabdh: Basically there are two flows where user is added API flow, upsert asset flow.

1. Upsert Asset flow
    
When upserting asset, each asset has owners information and the information is passed as e-mail. Compass by default added these e-mails to the Compass [Phantom User](https://github.com/odpf/compass/blob/main/docs/concepts/user.md) (user with UUID null but has e-mail). [code](https://github.com/odpf/compass/blob/79ca78b19d457371b90b31e90dee5378e37226af/internal/store/postgres/asset_repository.go#L555)

2. API flow

For every incoming API, compass check 2 headers, UUID header and E-mail header. UUID header is mandatory and E-mail header is optional. There will be 4 cases for this flow.
- UUID header exists > Check if UUID exists in compass > if true, request valid, if not, UpsertByEmail
- UUID header not exists, E-mail header exists > request rejected
- UUID header not exists, E-mail header not exists > request rejected

UpsertByEmail logic is: 
- if e-mail exist in compass and uuid is null, we update uuid with uuid that we get from header. 
- if e-mail exist in compass and uuid is not null, pass (this is the case that you need to solve, currently it throws generic error hence return 500 internal server error)
- if e-mail not exist in compass, insert uuid and e-mail.

scortier: If we want to check the type of error occuring because of same email and diff uuid, then the only way is to check it through `checkPostgresError` func which will identify the particular error, right now in query we were directly checking if conflict on email then update row as per need, which skips the 409 error conflict which we need to show instead of skipping it .
wdyt @mabdh ??

mabdh: If uuid and e-mail already exist and request is coming with same e-mail but different uuid, the query will return `error sql: no rows` (similar error when we update nothing)

You can check if error is sql no rows, if yes return a specific user error and in the handler, we can handle it with conflict.

internal/server/v1beta1/server.go
```diff
@@ -56,6 +56,9 @@ func (server *APIServer) validateUserInCtx(ctx context.Context) (string, error)
 		if errors.Is(err, user.ErrNoUserInformation) {
 			return "", status.Errorf(codes.InvalidArgument, err.Error())
 		}
+		if errors.Is(err, user.DuplicateRecordError{UUID: usr.UUID, Email: usr.Email}) {
```
mabdh, line 59: `user.DuplicateRecordError` is not a sentinel error, so better to use `As` rather than `Is` to check.

See https://go.dev/blog/go1.13-errors

mabdh (approved):

PR #146, opened 2022-06-07
feat(compass): add filter to get all types and move gettypes to postgres

mabdh:


StewartJingga (commented):
If count is 0, the field is not being sent. It is fine for grpc calls, but for HTTP(json) calls, the response will be not as informative.

One approach we can use is to tweak our grpc json marshaller to not omit empty [here](https://github.com/odpf/compass/blob/main/internal/server/server.go#L117).
example
```
gwmux := runtime.NewServeMux(
	runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{
		MarshalOptions: protojson.MarshalOptions{
			UseProtoNames: true,
			EmitUnpopulated: true,
		},
	}),
)
```

wdyt @mabdh ?

mabdh:
@StewartJingga yes that should be fine

StewartJingga (approved):
LGTM

PR #159, opened 2022-07-11
fix: rename homebrew-taps repo to homebrew-tap

GrayFlash:


ravisuhag:
@GrayFlash let's fix this in readme and installation guide as well.

Issue #162, opened 2022-08-01
feat: managing Asset via URN instead of ID


StewartJingga:
**Is your feature request related to a problem? Please describe.**
I am planning to have a job/pipeline that will be getting and deleting a single asset.
Right now compass has those APIs to get and delete, but it is using asset's ID.

Problem is, the job will not have the asset's ID available when running, instead it can generate asset's URN.

**Describe the solution you'd like**
- New API to retrieve an asset using its URN
- New API to delete an asset using its URN

**Describe alternatives you've considered**
- Use GetAllAssets API and filter the results

StewartJingga:
## Approach 1

Modify current APIs to also accept URN.
1. `GetAssetByID` - `[GET] /v1beta1/assets/{id}`
2. `DeleteAsset` - `[DELETE] /v1beta1/assets/{id}`

Compass internally will use the given `{id}` from route parameter as both, asset's `id` or `urn`.


## Approach 2

New API to use URN instead of ID
1. `GetAssetByURN` - `[GET] /v1beta1/assets/urn/{urn}`
2. `DeleteAssetByURN` - `[DELETE] /v1beta1/assets/urn/{urn}`


- **Approach 1** is overall simpler and quicker to implement.
- **Approach 2**'s HTTP endpoints cannot be RESTful as they all will conflict with current APIs using ID.

wdyt @bsushmith ?

bsushmith:
@StewartJingga I'm of a slightly different opinion here. 

**Approach 1**:
* We would alter the API definitions for this, and this can lead to confusion as it takes both values - either `{id}` or `{urn}`

**Approach 2**:
* The new APIs would be nice along with a deprecate tag on the APIs using `{id}`. We can remove them completely after a couple of versions. 


> Approach 2's HTTP endpoints cannot be RESTful as they all will conflict with current APIs using ID.

Can you elaborate on this? 
Both`[GET] /v1beta1/assets/{id}` and `[GET] /v1beta1/assets/urn/{urn}` will not have endpoint conflicts. Please CMIIW here.

StewartJingga:
> Can you elaborate on this?
> Both[GET] /v1beta1/assets/{id} and [GET] /v1beta1/assets/urn/{urn} will not have endpoint conflicts. Please CMIIW here.

@bsushmith They won't have conflict with the new endpoints that i'm proposing.
[...]

PR #165, opened 2022-08-16
fix: index assets table on urn column

sudo-suhas:
- Drop unique index on (urn, type, service) concurrently. This ensures
  that we don't take an exclusive lock on the table which would lock out
  selects, inserts, updates and deletes on the index's table.
- Create a unique index on urn column in the assets table.

Related to #162.


StewartJingga:
@sudo-suhas I might've missed something here, but what is the reason for splitting the migration? can they be in one use case?

sudo-suhas:
Already included the reason as a comment in one of the migration files:

```sql
-- By default, golang-migrate wraps multiple SQL statements in a transaction.
-- Dropping index concurrently is not allowed in a transaction. So the drop
-- statement needs to be the only statement in the migration.
```

StewartJingga (approved):

PR #166, opened 2022-08-16
feat: manage assets with URN for GET/DELETE

### LINKED ISSUE #162: feat: managing Asset via URN instead of ID

sudo-suhas:
- On the service layer, for get and delete, check if the given ID is a
  UUID and call corresponding method on repository (ex: GetByID or
  GetByURN). This introduces a limitation into the system where if an
  asset's URN is a UUID, the get and delete calls will fail. This
  limitation is considered acceptable in the short term and in the long
  term, the plan is to drop support for get or delete by ID. ID will be
  internal to Compass.
- On the repository layer for PG and ES, add get and delete by URN
  variants.
- Drop the Find method on asset repository and replace usages with
  GetWithURN.

BREAKING CHANGE: The interface for asset Repository and
DiscoveryRepository have been changed - methods have been renamed and
new methods have been added.

Closes #162

### DIFF
--- a/core/asset/asset.go
+++ b/core/asset/asset.go
@@ -13,12 +13,13 @@ type Repository interface {
 	GetAll(context.Context, Filter) ([]Asset, error)
 	GetCount(context.Context, Filter) (int, error)
 	GetByID(ctx context.Context, id string) (Asset, error)
-	Find(ctx context.Context, urn string, typ Type, service string) (Asset, error)
+	GetByURN(ctx context.Context, urn string) (Asset, error)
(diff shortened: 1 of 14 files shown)


StewartJingga (approved):
LGTM

PR #171, opened 2022-09-12
feat: new UpsertAsset api with replace strategy instead of patch/merge

### LINKED ISSUE #160: ability to overwrite asset instead of merge
As part of [UpdatePatchAsset](https://github.com/odpf/compass/blob/main/internal/server/v1beta1/asset.go#L215), we [patch](https://github.com/odpf/compass/blob/main/internal/server/v1beta1/asset.go#L237) and merge the asset before writing into DB. 

For the asset field `data` which contains the crux of asset information, it is impossible to overwrite in the current code. Only if the existing asset data is nil, it is possible to overwrite it. Otherwise, the existing `asset.data` is [merged](https://github.com/odpf/compass/blob/main/core/asset/patch.go#L100) with incoming changes before writing to DB.

Would like to have a feature flag, where the user can decide if they would like to always take it in the latest changes or merge it with existing values. 

**Describe the solution you'd like**
A feature flag to decide whether to merge or take it latest changes without merging.

**Describe alternatives you've considered**
I wanted to remove preview rows for bigquery tables in a project, and configured `meteor` recipe in such a way by having `max_preview_rows` value set as 0.  But since compass does only merge rather than overwrite, it didn't happen. I've tried to delete the asset and then insert the asset again. but since the `DeleteAsset` takes in asset id, it became cumbersome to delete all the assets using the id. 

**Additional context**
Am not sure about the process of finding asset `diff` here - https://github.com/odpf/compass/blob/main/internal/store/postgres/asset_repository.go#L301. By the time, the code flow reaches here, the asset is already merged with incoming changes. This diff is used to track asset versions in `asset_versions` table.
 Please correct me If I am wrong here, but shouldn't the diff be calculated with incoming changes rather than merged changes?

StewartJingga:
Close #160


internal/server/v1beta1/asset_test.go
```diff
@@ -280,6 +280,194 @@ func TestGetAssetByID(t *testing.T) {
...
+					Type:    "invalid type",
+				},
+			},
+			ExpectStatus: codes.InvalidArgument,
+		},
+		{
+			Description: "should return internal server error when upserting asset service failed",
+			Setup: func(ctx context.Context, as *mocks.AssetService) {
+				expectedErr := errors.New("unknown error")
+				as.EXPECT().UpsertAsset(ctx, mock.AnythingOfType("*asset.Asset"), mock.AnythingOfType("[]asset.LineageNode"), mock.AnythingOfType("[]asset.LineageNode")).Return("1234-5678", expectedErr)
```
sudo-suhas, line 393:
```suggestion
				as.EXPECT().UpsertAsset(ctx, mock.AnythingOfType("*asset.Asset"), mock.AnythingOfType("[]asset.LineageNode"), mock.AnythingOfType("[]asset.LineageNode")).Return("", expectedErr)
```

internal/server/v1beta1/asset.go
```diff
@@ -211,6 +211,58 @@ func (server *APIServer) GetAssetByVersion(ctx context.Context, req *compassv1be
...
+		return nil, err
+	}
+
+	baseAsset := req.GetAsset()
+	if baseAsset == nil {
+		return nil, status.Error(codes.InvalidArgument, "asset cannot be empty")
+	}
+
+	ast := server.buildAsset(baseAsset)
+	ast.UpdatedBy.ID = userID
```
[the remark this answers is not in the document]
StewartJingga:
```
type upsertRequest interface {
	GetAsset() *compassv1beta1.UpsertPatchAssetRequest_BaseAsset
}
```

We can't use this because `GetAsset()` response would be different for `UpsertAssetRequest`. Also both APIs are using different statsd metric name.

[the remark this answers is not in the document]
StewartJingga: @sudo-suhas updated here
https://github.com/odpf/compass/pull/171/commits/70347453366d12b35aded46d0e5e5ef1654957fd

internal/server/v1beta1/asset.go
```diff
@@ -211,6 +211,33 @@ func (server *APIServer) GetAssetByVersion(ctx context.Context, req *compassv1be
...
+
+	assetID, err := server.upsertAsset(
+		ctx,
+		ast,
+		"asset_upsert",
+		req.GetUpstreams(),
+		req.GetDownstreams(),
+	)
+
+	return &compassv1beta1.UpsertAssetResponse{
```
sudo-suhas, line 236: lets return `nil` value when `err != nil`.

internal/server/v1beta1/asset.go
```diff
@@ -234,44 +261,19 @@ func (server *APIServer) UpsertPatchAsset(ctx context.Context, req *compassv1bet
...
-		})
+	assetID, err := server.upsertAsset(
+		ctx,
+		ast,
+		"asset_upsert_patch",
+		req.GetUpstreams(),
+		req.GetDownstreams(),
+	)
 
 	return &compassv1beta1.UpsertPatchAssetResponse{
```
sudo-suhas, line 274: lets return `nil` value when `err != nil` here as well.

StewartJingga: updated

Makefile
```diff
@@ -1,7 +1,7 @@
 NAME="github.com/odpf/compass"
 VERSION=$(shell git describe --always --tags 2>/dev/null)
 COVERFILE="/tmp/compass.coverprofile"
-PROTON_COMMIT := "9eaca223dfd5ca0297abdb593fa73e8522d8ef62"
+PROTON_COMMIT := "9d1d2af4192b0b31e95d417bbd17752825776a08"
```
sudo-suhas, line 4:
```suggestion
PROTON_COMMIT := "ab6d8cf86c6c86c8eba0ed7dde79c36eb2acdbf5"
```

The commit on master - https://github.com/odpf/proton/commit/ab6d8cf86c6c86c8eba0ed7dde79c36eb2acdbf5

StewartJingga: updated using latest proton commit https://github.com/odpf/proton/commit/bfab84cdbc61ffa8725bcef7e44635885488bf24

sudo-suhas (approved):

sudo-suhas:
@StewartJingga shall we merge in these changes?

PR #176, opened 2022-10-06
feat: allow feature_table as an asset type

sudo-suhas:
Related to https://github.com/odpf/meteor/pull/417

### DIFF
--- a/core/asset/type.go
+++ b/core/asset/type.go
@@ -1,10 +1,11 @@
 package asset
 
 const (
-	TypeTable     Type = "table"
-	TypeJob       Type = "job"
-	TypeDashboard Type = "dashboard"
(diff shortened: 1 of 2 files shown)


StewartJingga:
LGTM, just a small concern if `ml_feature_table` is a good type name? cc: @sudo-suhas @ravisuhag

sudo-suhas:
Do you have a suggestion @StewartJingga? In the PR for adding E2E lineage entities (https://github.com/odpf/proton/pull/193), we had discussed calling it `FeatureTable` instead of `MLFeatureTable` but chose to keep it as `MLFeatureTable` because 'Feature' is an overloaded word that is not clear enough.

StewartJingga (approved):
LGTM

PR #178, opened 2022-10-12
refactor: stop usage of deprecated fields in LineageNode

sudo-suhas:
- Stop usage of v1beta1.LineageNode.{type, service}. The URN is now sufficient to uniquely identify the node.
- Modify interfaces and service to stop usage of core/asset.LineageNode, just use the URN.
  - core/asset.Service struct.
  - core/asset/LineageRepository interface.
- Remove the type core/asset.LineageEdge.

BREAKING CHANGES:
- Types modified - core/asset.Service, core/asset/LineageRepository.
- Type removed - core/asset.LineageEdge.

PR #180, opened 2022-10-14
feat: include latest probe in lineage response

sudo-suhas:
- Return node attributes in Lineage response for GetGraph call. Node
  attributes include latest probe for asset nodes in the Graph. This
  enables showing live lineage.
- Use context for gRPC dial and compass service handler in server.
- Add support for fetching probes with filters from asset repository.
  Supported filters:
  - AssetURNs - zero or more asset URNs.
  - MaxRows - Number of rows for each Asset URN ordered by latest first.
  - OlderThan,NewerThan - Data range.
- Fix some lint errors.

Done: _Should be merged after #178 (would be rebased on main)_


StewartJingga:
LGTM, but found an issue that may or may not be solved in this PR
https://github.com/odpf/compass/issues/182

It is out of scope for sure, but can this be easily solved in this PR?
@sudo-suhas

sudo-suhas:
Created a different PR to fix that issue - https://github.com/odpf/compass/pull/183

This way git history is cleaner and we are not including unrelated change in this PR.

StewartJingga (approved):
LGTM

PR #183, opened 2022-10-27
fix: require timestamp for adding asset probe

### LINKED ISSUE #182: fix: probe's timestamp value is set to default time value instead of empty if not defined
**Describe the bug**
When creating a new probe via `[POST] /v1beta1/assets/{asset_urn}/probes`, `timestamp` field is not required hence we can submit without it.

This causes confusion, especially when fetching the latest probe via `[GET] /v1beta1/lineage/{asset_urn}`.
```
{
    "data": [
        {
            "source": "sample-urn",
            "target": "sample-urn-2"
        }
    ],
    "node_attrs": {
        "sample-urn": {
            "probes": {
                "latest": {
                    "id": "239c0a6a-821c-4312-bcb5-f0b112435408",
                    "asset_urn": "sample-urn",
                    "status": "STARTING",
                    "status_reason": "Test Reason",
                    "metadata": {
                        "is_test": true,
                        "test_id": 1,
                        "type": "test"
                    },
                    "timestamp": "1970-01-01T00:00:00Z",
                    "created_at": "2022-10-27T09:56:38.339542Z"
                }
            }
        }
    }
}
```

From the response, we can see that `node_attrs["sample_urn"].probes.latest.timestamp`  is `1970-01-01T00:00:00Z`, which is a default value for time in Golang.

Client will not be sure which timestamp to use, `timestamp` or `created_at`.
If client wants to use `created_at` if timestamp is not defined, it will have trouble figuring out if `timestamp` is empty or not because Compass returns default time value instead of `null` or not all.

**To Reproduce**
1. Create an asset
3. Create a probe for that asset without timestamp field (`[POST] /v1beta1/assets/{asset_urn}/probes`)
4. Fetch lineage of the asset
5. See timestamp is set as default time value

**Expected behavior**
Ideally Compass should either
1. make `timestamp` as a required field when creating probe, or
2. set `timestamp` as `created_at` if it not submitted by client

sudo-suhas:
Fixes #182

### DIFF
--- a/internal/server/v1beta1/asset.go
+++ b/internal/server/v1beta1/asset.go
@@ -318,6 +318,9 @@ func (server *APIServer) CreateAssetProbe(ctx context.Context, req *compassv1bet
 	if req.Probe.Status == "" {
 		return nil, status.Error(codes.InvalidArgument, "Status is required")
 	}
+	if !req.Probe.Timestamp.IsValid() {
+		return nil, status.Error(codes.InvalidArgument, "Timestamp is required")
+	}
 
 	probe := asset.Probe{
 		Status:       req.Probe.Status,
--- a/internal/server/v1beta1/asset_test.go
+++ b/internal/server/v1beta1/asset_test.go
@@ -1120,11 +1120,23 @@ func TestCreateAssetProbe(t *testing.T) {
 
 	var testCases = []testCase{
 		{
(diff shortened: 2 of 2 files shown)


StewartJingga (approved):
LGTM

Ghjkvvgj:
- [ ] - @pyadav

PR #189, opened 2022-11-30
feat: serve grpc on different port

StewartJingga:
- Use [odpf/salt/mux](https://github.com/odpf/salt/pull/63) package to server grpc on different port
- Fix backward compatibility issues when updating salt version on `salt/term` package

Relates to https://github.com/odpf/salt/pull/63


internal/server/server.go
```diff
@@ -127,56 +128,31 @@ func Serve(
...
-		Handler:      grpcHandlerFunc(grpcServer, baseMux),
-		Addr:         address,
-		ReadTimeout:  5 * time.Second,
-		WriteTimeout: 10 * time.Second,
-		IdleTimeout:  120 * time.Second,
+	logger.Info("server started")
+	if err := mux.Serve(
+		ctx,
+		mux.WithHTTPTarget(config.addr(), &http.Server{
+			Handler:      grpcHandlerFunc(grpcServer, httpMux),
```
sudo-suhas, line 138: Why are we using `grpcHandlerFunc` here? We should be able to directly pass `gwmux` here.

internal/server/server.go
```diff
@@ -127,56 +128,31 @@ func Serve(
...
+	httpMux := http.NewServeMux()
+	httpMux.Handle("/", gwmux)
 
-	httpServer := &http.Server{
-		Handler:      grpcHandlerFunc(grpcServer, baseMux),
-		Addr:         address,
-		ReadTimeout:  5 * time.Second,
-		WriteTimeout: 10 * time.Second,
-		IdleTimeout:  120 * time.Second,
+	logger.Info("server started")
```
sudo-suhas, line 134:
```suggestion
	logger.Info("Starting server", "http_port", config.addr(), "grpc_port", config.grpcAddr())
```

internal/server/server.go
```diff
@@ -127,56 +128,31 @@ func Serve(
...
+		ctx,
+		mux.WithHTTPTarget(config.addr(), &http.Server{
+			Handler:      grpcHandlerFunc(grpcServer, httpMux),
+			ReadTimeout:  5 * time.Second,
+			WriteTimeout: 10 * time.Second,
+			IdleTimeout:  120 * time.Second,
+		}),
+		mux.WithGRPCTarget(config.grpcAddr(), grpcServer),
+		mux.WithGracePeriod(5*time.Second),
+	); err != ctx.Err() {
```
sudo-suhas, line 145:
```suggestion
	); !errors.Is(err, context.Canceled) {
```

internal/server/server.go
```diff
@@ -196,7 +172,7 @@ func makeHeaderMatcher(c Config) func(key string) (string, bool) {
 	}
 }
 
-// grpcHandlerFunc routes http1 calls to baseMux and http2 with grpc header to grpcServer.
+// grpcHandlerFunc routes http1 calls to httpMux and http2 with grpc header to grpcServer.
```
sudo-suhas, line 175: This function can be removed.

internal/server/server.go
```diff
@@ -127,56 +128,31 @@ func Serve(
...
-			logger.Warn("stopping http server...")
-			if err := httpServer.Shutdown(context.Background()); err != nil {
-				logger.Error("HTTP server Shutdown", "err", err)
-			}
-		}
-
-		if grpcServer != nil {
-			logger.Warn("stopping grpc server...")
-			grpcServer.GracefulStop()
+	if pgClient != nil {
```
sudo-suhas, line 149: Might be better to keep this as a deferred function so that it runs even if there is a panic.

sudo-suhas (changes requested):

sudo-suhas (approved):
LGTM

PR #192, opened 2022-12-14
chore: increase grpc max call send & receive msg size

bsushmith:
* Increase the grpc max call send & receive msg size to `32MB` from default `4MB`.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -30,22 +30,20 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
-const (
-	GRPCMaxClientSendSize = 32 << 20
-)
-
 type Config struct {
-	Host     string `mapstructure:"host" default:"0.0.0.0"`
-	Port     int    `mapstructure:"port" default:"8080"`
-	GRPCPort int    `mapstructure:"grpc_port" default:"8081"`
-	BaseUrl  string `mapstructure:"baseurl" default:"localhost:8080"`
+	Host    string `mapstructure:"host" default:"0.0.0.0"`
+	Port    int    `mapstructure:"port" default:"8080"`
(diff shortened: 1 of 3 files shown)


internal/server/server.go
```diff
@@ -100,7 +104,14 @@ func Serve(
 
 	headerMatcher := makeHeaderMatcher(config)
 
-	grpcConn, err := grpc.DialContext(grpcDialCtx, config.grpcAddr(), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	grpcConn, err := grpc.DialContext(
+		grpcDialCtx,
+		config.grpcAddr(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(GRPCMaxClientSendSize),
```
StewartJingga, line 112: Do we need to increase RecvMsg as well? All of our insert APIs do not seem to need it.
wdyt?

internal/server/server.go
```diff
@@ -30,6 +30,10 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+const (
+	GRPCMaxClientSendSize = 32 << 20
```
StewartJingga, line 34: should we make it configurable as well ? wdyt?

bsushmith: Yes. I think we should be able to configure it also if needed.

internal/server/server.go
bsushmith: Not sure about this. 
Increased the `RecvMsg` as well, as asset payload can be huge - but haven't faced this scenario yet.

internal/server/server.go
bsushmith: made both values to be configurable

sudo-suhas (approved):

PR #195, opened 2023-01-16
fix: Patch asset's lineage

sudo-suhas:
- For the UpsertPatchAsset RPC, when no upstreams and downstreams are specified in the request, skip overwriting the asset's lineage unless OverwriteLineage flag is set to true.
- Bump up the proton commit in Makefile to pull in changes from https://github.com/odpf/proton/pull/238 + generate protos.


StewartJingga (approved):
LGTM

PR #204, opened 2023-02-07
feat!: add cli commands

Chief-Rishab:
Add/Update CLI commands
- [x] Style: update core commands according to the [salt](https://github.com/odpf/salt) configurations
- [x] Reference 
- [x] Shell auto completion, 
- [x] Environment
- [x] Config init (Refactor `configs` cmd to support `config init` and `config list`)
- [x] Version
- [x] Refactor `compass serve` and `compass migrate` command to `compass server <subcommands>`
- [x] Asset: getalltypes, stargazers, version history, get asset by version
- [ ] Tags: create, edit, delete, list tag asset and create, edit, delete, list tag template
- [x] Current user: get user starred assets, star asset, unstar asset, get my discussions
- [ ] Discussion: edit discussion, view, create, update, delete and list comments
- [ ] Search: Suggest
- [x] Make file to auto-generate docs, fix version flags

BREAKING CHANGES: `serve` and `migrate` command changed to `server start`, `server migrate`, 
`configs` changed to `config init` and `config list`


Chief-Rishab:
@ravisuhag can you start reviewing the changes here? The rest I will push in a different branch

ravisuhag (approved):

sudo-suhas:
Is it possible to split this PR into smaller changesets? Additionally, can we call out the breaking changes and the rationale for why we think they are necessary?

ravisuhag:
@sudo-suhas Only breaking one i see is the server commands which are in alignment with rest of the products. If possible let's review this one. We can keep these short going forward.

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+	cmd.Flags().StringVarP(&types, "types", "t", "", "filter by types")
+	cmd.Flags().StringVarP(&services, "services", "s", "", "filter by services")
+	cmd.Flags().StringVarP(&data, "data", "d", "", "filter by field in asset.data")
+	cmd.Flags().StringVar(&q, "query", "", "filter by specific query")
+	cmd.Flags().StringVar(&q_fields, "query_fields", "", "filter by query field")
+
+	return cmd
+}
+
+func makeGetAllTypesRequest(types, services, data, q, q_fields string) *compassv1beta1.GetAllTypesRequest {
```
kushsharma, line 324: Nothing specific to this code but I wonder why we have filters on a simple list API of asset types. @sudo-suhas can you help me understand the usecase?

sudo-suhas:
There are a large number of changes in the PR making it difficult to review. Therefore requesting again to split the PR into smaller changesets.

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+	cmd.Flags().Uint32Var(&size, "size", 20, "Size of each page")
+	cmd.Flags().Uint32Var(&page, "page", 0, "Number of pages")
+
+	return cmd
+}
+
+func starAssetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "star <id>",
+		Short: "star an asset by id for current user",
```
kushsharma, line 386: I guess this should work with id or urn?

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+			defer cancel()
+
+			assetID := args[0]
+			ctx := client.SetMetadata(cmd.Context())
+			_, err = clnt.StarAsset(ctx, &compassv1beta1.StarAssetRequest{
+				AssetId: assetID,
+			})
+			if err != nil {
+				return err
+			}
```
kushsharma, line 407: We can combine statements like these to single if err check. For eg.
```go
if _, err := clnt.StarAsset(ctx, &compassv1beta1.StarAssetRequest{
    AssetId: assetID,
}); err != nil {
    return err
}
```

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func unstarAssetCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unstar <id>",
```
kushsharma, line 421: As most of the added commands expect a positional argument, I would advice we can use a validation check as
```go
Args: cobra.MinimumNArgs(1),
```

ravisuhag:
@Chief-Rishab lets split docs and cli changes into different PRs.

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+			} else {
+				fmt.Println(term.Bluef(prettyPrint(res.GetData())))
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&json, "out", "o", "table", "flag to control output viewing, for json `-o json`")
+	cmd.Flags().Uint32Var(&size, "size", 20, "Size of each page")
```
kushsharma, line 504: We can make a const for all of the command defaults like `size` in this case.

cli/assets.go
```diff
@@ -256,3 +270,318 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+			spinner.Stop()
+
+			fmt.Println(term.Bluef(prettyPrint(res.GetData())))
+
+			return nil
+		},
+	}
+
+	cmd.Flags().Uint32Var(&size, "size", 20, "Size of each page")
+	cmd.Flags().Uint32Var(&page, "page", 0, "Number of pages")
```
kushsharma, line 546: It's not really the number of pages, its the page number offset of results. So if we have total 60 results, default 0 will give us the first page of 3(60/20). Page 1 will give rows from 20-40.

cli/config.go
```diff
@@ -1,36 +1,84 @@
...
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v2"
 )
 
 const configFlag = "config"
 
-func cmdShowConfigs() *cobra.Command {
+var cliConfig *Config
+
+func configCommand() *cobra.Command {
```
kushsharma, line 21: Do we have a single file for client and server configs? I see we are loading client configurations but the removed code was reading server configurations. Although I believe instead of managing two files, we could have used one for both.

cli/root.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"github.com/MakeNowJust/heredoc"
+	"github.com/odpf/compass/internal/client"
+	"github.com/odpf/salt/cmdx"
+	"github.com/spf13/cobra"
+)
+
+func New(cfg *Config) *cobra.Command {
+	cliConfig = cfg
```
kushsharma, line 11: I see we are using a global variable for config but also in the `server start` command loading configs again?

cli/server.go
```diff
@@ -28,27 +28,66 @@ var (
...
+	c := &cobra.Command{
+		Use:   "migrate",
+		Short: "Run storage migration",
+		Example: heredoc.Doc(`
+			$ compass server migrate
+		`),
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cfg, err := loadConfig(cmd)
+			ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
kushsharma, line 79: `cmd.Context()`?

kushsharma (changes requested):

cli/config.go
Chief-Rishab: Yes we have a single file for both

cli/assets.go
Chief-Rishab: The star asset proto definition had only AssetID, also tried with the http client, wasnt possible to star with urn currently

cli/assets.go
Chief-Rishab: was getting this syntax error with the suggested change 
```expected operand, found 'return' syntax```

Chief-Rishab:
@sudo-suhas have kept the cli changes in this branch, docs in another. Can you review it once

PR #209, opened 2023-02-21
fix: store, retrieve asset.url

sudo-suhas:
- Add column to assets table for URL.
- Store the asset's URL field in DB as part of insert/upsert.
- Retrieve the asset's URL field from DB for fetch APIs.
- Refactor UPDATE statement to use squirrel for easier maintenance.


internal/store/postgres/asset_repository.go
```diff
@@ -775,6 +776,7 @@ func (r *AssetRepository) getAssetSQL() sq.SelectBuilder {
 		a.service as service,
 		a.description as description,
 		a.data as data,
+		COALESCE(a.url, '') as url,
```
haveiss, line 779: how this compared to setting url default to '' when altering table?

haveiss (approved):

internal/store/postgres/asset_repository.go
sudo-suhas: Altering a table with default value used to take a lock for updating all rows (which would block reads and writes). This looks to have been fixed in PG 11 but I wasn't aware of that while making these changes.

PR #211, opened 2023-02-22
feat!: update cli commands

Chief-Rishab:
Add/Update CLI commands
- [x] Style: update core commands according to the [salt](https://github.com/odpf/salt) configurations
- [x] Reference 
- [x] Shell auto completion, 
- [x] Environment
- [x] Config init (Refactor `configs` cmd to support `config init` and `config list`)
- [x] Version
- [x] Refactor `compass serve` and `compass migrate` command to `compass server <subcommands>`
- [x] Asset: getalltypes, stargazers, version history, get asset by version
- [x] Current user: get user starred assets, star asset, unstar asset, get my discussions
- [x] Make file to auto-generate docs, fix version flags
- [x] Fix the client config over-riding from config flag ( currently the client config from root is not getting over-ridden) 

BREAKING CHANGES: `serve` and `migrate` command changed to `server start`, `server migrate`, 
`configs` changed to `config init` and `config list`


ravisuhag (approved):

ravisuhag:
@sudo-suhas @StewartJingga can you take a look at this. If looks good we can merge this. This follows similar guidelines as we are using in Guardian for CLI.

sudo-suhas:
@ravisuhag I am on leave till 24th, will check once I am back.

cli/assets.go
```diff
@@ -256,3 +279,325 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+		`),
+		Args: cobra.NoArgs,
+		Annotations: map[string]string{
+			"action:core": "true",
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			spinner := printer.Spin("")
+			defer spinner.Stop()
+
+			clnt, cancel, err := client.Create(cmd.Context())
```
sudo-suhas, line 300: Docs for `cmd.Context()`.
```
// Context returns underlying command context. If command wasn't
// executed with ExecuteContext Context returns Background context.
```

We aren't executing the command with `ExecuteContext` currently, please fix.

cli/root.go
```diff
@@ -0,0 +1,62 @@
...
+	rootCmd.AddCommand(cmdx.SetCompletionCmd("compass"))
+	rootCmd.AddCommand(cmdx.SetRefCmd(rootCmd))
+	rootCmd.AddCommand(cmdx.SetHelpTopicCmd("environment", envHelp))
+	cmdx.SetHelp(rootCmd)
+
+	if cliConfig.Client.ServerHeaderKeyUUID == "" {
+		cliConfig.Client.ServerHeaderKeyUUID = cliConfig.Service.Identity.HeaderKeyUUID
+	}
+
+	rootCmd.PersistentFlags().StringP(configFlag, "c", "", "Override config file")
```
sudo-suhas, line 59: We need to respect the override in all commands. It is currently only supported in server start and migrate commands.

cli/server.go
```diff
@@ -28,51 +28,101 @@ var (
 	Version string
 )
 
-func cmdServe() *cobra.Command {
-	return &cobra.Command{
-		Use:     "serve",
-		Short:   "Serve gRPC & HTTP service",
-		Long:    heredoc.Doc(`Serve gRPC & HTTP on a port defined in PORT env var.`),
-		Aliases: []string{"server", "start"},
+func serverCmd() *cobra.Command {
```
sudo-suhas, line 31: _Currently not able to test these since the config loading is broken._

cli/server.go
```diff
@@ -28,51 +28,101 @@ var (
...
 		return err
 	}
-	statsdReporter, err := statsd.Init(logger, config.StatsD)
+	statsdReporter, err := statsd.Init(logger, cliConfig.StatsD)
 	if err != nil {
 		return err
 	}
 
-	esClient, err := initElasticsearch(logger, config.Elasticsearch)
+	esClient, err := initElasticsearch(logger)
```
sudo-suhas, line 120: Pass in the configs explicitly.

cli/server.go
```diff
@@ -28,51 +28,101 @@ var (
...
+			return runMigrations(ctx)
 		},
 	}
+	c.Flags().StringVarP(&configFile, "config", "c", "config.yaml", "Config file path")
+
+	return c
 }
 
-func runServer(config Config) error {
+func runServer() error {
```
sudo-suhas, line 104: Undo this change please.

cli/version.go
```diff
@@ -0,0 +1,28 @@
...
+
+// VersionCmd prints the version of the binary
+func versionCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:     "version",
+		Aliases: []string{"v"},
+		Short:   "Print version information",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if Version == "" {
+				fmt.Println(term.Yellow("Version history not available"))
```
sudo-suhas, line 19: Any reason we are printing 'Version history'? Shouldn't it just be 'Version'?

main.go
```diff
@@ -1,7 +1,60 @@
...
 func main() {
-	cli.Execute()
+	cliConfig, err := cli.LoadConfig()
+	if err != nil {
+		cliConfig = &cli.Config{}
+	}
+
+	rootCmd := cli.New(cliConfig)
+
+	if cmd, err := rootCmd.ExecuteC(); err != nil {
```
sudo-suhas, line 24: Can we run with `ExecuteContextC` instead? We can pass in a context created with `signal.NotifyContext`.

sudo-suhas (changes requested):

main.go
```diff
@@ -1,7 +1,60 @@
...
+		printError(err)
+
+		cmdErr := strings.HasPrefix(err.Error(), "unknown command")
+		flagErr := strings.HasPrefix(err.Error(), "unknown flag")
+		sflagErr := strings.HasPrefix(err.Error(), "unknown shorthand flag")
+
+		if cmdErr || flagErr || sflagErr {
```
sudo-suhas, line 31: Can you double-check the handling for this? Usage is being printed twice currently:

```
$ ./compass config --test
Usage:  compass config <command> [flags]

Available commands:
  init
  list

unknown flag: --test

Usage:  compass config <command> [flags]

Available commands:
  init
  list
```

cli/assets.go
```diff
@@ -134,6 +154,7 @@ func viewAssetByIDCommand() *cobra.Command {
 		Example: heredoc.Doc(`
 			$ compass asset view <id>
```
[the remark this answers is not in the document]
Chief-Rishab: will include the changes in main and update accordingly

cli/config.go
```diff
@@ -3,34 +3,84 @@ package cli
...
+				return err
+			}
+
+			fmt.Printf("config created: %v\n", cfg.File())
+			return nil
+		},
+	}
+}
+
+func configListCommand() *cobra.Command {
```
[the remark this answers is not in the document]
Chief-Rishab: @sudo-suhas we can load the config from custom file location with a `-c` or `--config` flag.
So for loading the file from the current directory you can use `compass config list -c ./compass.yml` 

Otherwise, by default it will try to load the file which we made from config init command

cli/server.go
Chief-Rishab: can you once again try with `--config` flag to use the file from current directory. Alternatively you can use the config file created after init command.

cli/assets.go
```diff
@@ -256,3 +279,325 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+
+	cmd.Flags().Uint32Var(&size, "size", pageSize, "Size of each page")
+	cmd.Flags().Uint32Var(&page, "page", pageOffset, "Page number offset (start from 0)")
+
+	return cmd
+}
+
+func viewAssetByVersionCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "version <id> <version>",
```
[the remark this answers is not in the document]
Chief-Rishab: sure, will add changes done in main recently

cli/assets.go
```diff
@@ -256,3 +279,325 @@ func deleteAssetByIDCommand() *cobra.Command {
...
+	cmd.Flags().StringVarP(&data, "data", "d", "", "filter by field in asset.data")
+	cmd.Flags().StringVar(&q, "query", "", "filter by specific query")
+	cmd.Flags().StringVar(&q_fields, "query_fields", "", "filter by query field")
+
+	return cmd
+}
+
+func makeGetAllTypesRequest(types, services, data, q, q_fields string) *compassv1beta1.GetAllTypesRequest {
+	newReq := &compassv1beta1.GetAllTypesRequest{}
+	if types != "" {
```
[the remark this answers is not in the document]
Chief-Rishab: Was already there in the current CLI code, will change if it's required

cli/config.go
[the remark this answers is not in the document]
Chief-Rishab: Edit: understood what you meant by current flow, will add the current directory to search for the config file, before looking for default location

cli/root.go
Chief-Rishab: @sudo-suhas is it there a reason why config overriding needs to be supported apart from any of these commands, wanted to know. We will be overriding the configs only during server start and migrate right?

cli/assets.go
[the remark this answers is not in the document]
sudo-suhas: Lets refactor where it makes sense.

cli/config.go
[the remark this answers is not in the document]
sudo-suhas: Can we please not break backward compatibility here? Not sure I see the value in doing so. Supporting a new default is okay but breaking existing functionality for insignificant gains should be avoided.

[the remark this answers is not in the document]
sudo-suhas: > Edit: understood what you meant by current flow, will add the current directory to search for the config file, before looking for default location

if this is the approach we are taking, we are partially in agreement. Please look into the previous behaviour and ensure the same is supported.

cli/root.go
sudo-suhas: Otherwise what is the use of having the persistent flag? Also, don't the commands for client functionality also use the config? Any reason why we don't want to support the override functionality there despite declaring a persistent flag?

cli/server.go
sudo-suhas: will test once the current set of review comments are addressed.

cli/config.go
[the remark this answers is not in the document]
Chief-Rishab: yes, I am adding the existing flow for loading  from root as well

cli/server.go
sudo-suhas: not sure if you realised but https://github.com/odpf/compass/pull/211#discussion_r1116555739 means that loading config is also broken.

cli/root.go
Chief-Rishab: I just wanted to understand how is it being used. Got your point

cli/config.go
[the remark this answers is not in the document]
Chief-Rishab: @sudo-suhas I have restored the option to load from the flags, or root. If files aren't available then I am loading from the file created from the init command defined in cmdx/config in salt. (Option 1)

An alternative option is to edit the cmdx package to find the file in the root. (Option 2) 
Or we can just copy the Loading part with configdir defined in here to compass (Option 3) https://github.com/odpf/salt/blob/main/cmdx/config.go#L83.

Let me know if the current changes are fine, or should we change it?

cli/config.go
```diff
@@ -56,27 +106,25 @@ type Config struct {
...
-	} else {
-		opts = append(opts,
-			config.WithPath("./"),
-			config.WithName("compass"),
-			config.WithEnvKeyReplacer(".", "_"),
-			config.WithEnvPrefix("COMPASS"),
-		)
-	}
+func LoadConfig() (*Config, error) {
+	var config Config
```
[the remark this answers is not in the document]
Chief-Rishab: @sudo-suhas I have restored the option to load from the flags, or root.(which is currently there) If these files aren't available then I am loading from the file created from the init command defined in cmdx/config in salt. (Option 1)

An alternative option is to edit the cmdx package in salt to find the file in the root as well. (Option 2) 
Or we can just copy the Loading part with configdir defined in here to compass (Option 3) https://github.com/odpf/salt/blob/main/cmdx/config.go#L83.

Let me know if the current changes are fine, or should we change it?

cli/config.go
```diff
@@ -65,18 +115,29 @@ func loadConfig(cmd *cobra.Command) (Config, error) {
 	} else {
 		opts = append(opts,
 			config.WithPath("./"),
-			config.WithName("compass"),
+			config.WithName("compass.yaml"),
```
Chief-Rishab, line 118: @sudo-suhas the options `WithName("compass")` gives error `error in reading config` when the ./compass.yaml is missing.
Instead it shoould return the file not found error. Currently it is reading the compass executable and through the error. Can we rename the file to `.config` instead. And the line 118
`config.WithName("compass.yaml"),` will be changed to `config.WithName(".config"),`. Will remove the extension .yaml after that?

Chief-Rishab, line 118: @sudo-suhas laoding the options `WithName("compass")` gives an error `error in reading config: yaml` when the conifg file `compass.yaml` is missing in the root folder.

Instead it should be giving an error, `file not found`. Currently it is trying to use the compass executable while trying to read the configs. 

I also tried to add the `WithType("yaml")` but it still gives the same error. Can we rename the `compass.yaml` to something like `.conifg.yaml`. 

I need the correct error to be handled for searching for file made from the init command, which will be called only when the config file is missing in the root.

The extention "yaml" in line 118 will be discarded once we go ahead with something which throws the right error.

sudo-suhas: I am not clear on why we care about the exact error message being returned. We were anyway proceeding with defaults when the config file was not found.

cli/config.go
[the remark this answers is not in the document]
sudo-suhas: Option 2 would be preferable. We can add new [`ConfigLoaderOpts`](https://pkg.go.dev/github.com/odpf/salt@v0.2.4/cmdx#ConfigLoaderOpts) to enable the functionality needed.

cli/config.go
```diff
@@ -5,37 +5,81 @@ import (
...
-			cfg, err := loadConfig(cmd)
-			if err != nil {
-				fmt.Printf("failed to read configs: %v\n", err)
-				os.Exit(1)
+		Use:   "init",
+		Short: "Initialize a new sevrer and client configuration",
+		Example: heredoc.Doc(`
+			$ compass config init
+		`),
+		// SilencePersistentFlag: true,
```
kushsharma, line 45: clean this up?

cli/config.go
```diff
@@ -5,37 +5,81 @@ import (
...
+		Short: "List server and client configuration settings",
+		Example: heredoc.Doc(`
+			$ compass config list
+		`),
+		Annotations: map[string]string{
+			"group": "core",
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			_ = yaml.NewEncoder(os.Stdout).Encode(*cfg)
+			return nil
```
kushsharma, line 74: `return yaml.NewEncoder(os.Stdout).Encode(*cfg)` ?

cli/config.go
```diff
@@ -56,27 +100,41 @@ type Config struct {
...
 		if errors.As(err, &config.ConfigFileNotFoundError{}) {
-			return cfg, nil
+			return &cfg, ErrConfigNotFound
 		}
-		return cfg, err
+		return &cfg, err
 	}
-	return cfg, nil
+	return &cfg, nil
+}
```
kushsharma, line 133: The current implementation looks like we give lower priority to the config file in the current directory, I feel we should do the opposite, and try to load from the current first if fails, we can fall back to the rest of the pre-defined directories.

cli/server.go
```diff
@@ -28,30 +28,64 @@ var (
...
+	cmd.AddCommand(
+		serverStartCommand(cfg),
+		serverMigrateCommand(cfg),
+	)
+
+	return cmd
+}
+
+func serverStartCommand(cfg *Config) *cobra.Command {
+
```
kushsharma, line 54: nit: I see couple of unnecessary new lines, for example in `serverMigrateCommand`, `func New(cliConfig *Config) *cobra.Comman`, and here

kushsharma (changes requested):

kushsharma:
LGTM

PR #213, opened 2023-02-23
feat: add asset data attributes to lineage response

### LINKED ISSUE #206: add asset data.properties.attributes in lineage response
In meteor we support enriching asset by adding user configurable(through recipe) key-value pairs to asset metadata. This is generally stored in `asset.data.properties.attributes` field. Add support to return back these data points as part of lineage API.

This can help the caller to do additional actions if necessary on the lineage response, based on the attributes.

bsushmith:
- [x] Update proto
- [x] code changes
- [ ] test cases

Proton PR - https://github.com/odpf/proton/pull/245

Closes #206


go.mod
```diff
@@ -20,24 +23,29 @@ require (
...
 	github.com/jackc/pgx/v4 v4.10.1
 	github.com/jmoiron/sqlx v1.3.5
 	github.com/lib/pq v1.10.2
+	github.com/lyft/protoc-gen-star v0.6.1 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/newrelic/go-agent/v3 v3.15.2
 	github.com/newrelic/go-agent/v3/integrations/nrelasticsearch-v7 v1.0.1
 	github.com/newrelic/go-agent/v3/integrations/nrgrpc v1.3.1
-	github.com/odpf/salt v0.2.5-0.20221130085531-51c81815f7d6
+	github.com/odpf/salt v0.2.1
```
sudo-suhas, line 40: This change should not be done. Can you dougle check the changes to go.mod?

internal/store/postgres/asset_repository.go
```diff
@@ -431,6 +431,29 @@ func (r *AssetRepository) GetProbesWithFilter(ctx context.Context, flt asset.Pro
 	return results, nil
 }
 
+func (r *AssetRepository) GetAssetDataAttributes(ctx context.Context, urns []string) (map[string]asset.Asset, error) {
+	stmt := sq.Select(`
+			id, urn, name, type, data->'attributes' as data
```
sudo-suhas, line 436: Can you elaborate on the use case for selectively fetching `data.attributes`? Not 100% sure if it makes sense outside of particular use cases.

internal/store/postgres/asset_repository.go
```diff
@@ -431,6 +431,29 @@ func (r *AssetRepository) GetProbesWithFilter(ctx context.Context, flt asset.Pro
...
+	`).From("assets").Where(sq.Eq{"urn": urns})
+
+	query, args, err := stmt.PlaceholderFormat(sq.Dollar).ToSql()
+	if err != nil {
+		return nil, fmt.Errorf("get assets with urns: build query: %w", err)
+	}
+
+	var assets []asset.Asset
```
sudo-suhas, line 444: Within the repository layer, we use the `AssetModel` to decode DB representation. Can we do the same here?

core/asset/service.go
```diff
@@ -138,9 +138,14 @@ func (s *Service) GetLineage(ctx context.Context, urn string, query LineageQuery
 		return Lineage{}, fmt.Errorf("get lineage: get latest probes: %w", err)
 	}
 
+	assetDataAttributes, err := s.assetRepository.GetAssetDataAttributes(ctx, urns.list())
```
sudo-suhas, line 141: I think this would need to be explicitly requested. Asset probes are very much within the lineage domain so it makes sense to return them by default.

core/asset/service.go
```diff
@@ -164,15 +169,22 @@ func isValidUUID(u string) bool {
...
 	nodeAttrs := make(map[string]NodeAttributes, len(assetProbes))
 	for urn, probes := range assetProbes {
-		if len(probes) == 0 {
-			continue
+		var probesInfo ProbesInfo
+		if len(probes) > 0 {
+			probesInfo = ProbesInfo{Latest: probes[0]}
+		}
+
+		attrs := make(map[string]interface{})
```
sudo-suhas, line 180: This would mean that attributes are included only if probes are present for the asset. Would recommend adding test cases to catch such issues.

sudo-suhas (changes requested):

core/asset/service.go
bsushmith: This was something that I was confused about how to make these changes... Will add a request param for this. 👍

internal/store/postgres/asset_repository.go
bsushmith: Yes.  In our Frontend UI which calls the lineage API - we are planning to show some filters based on attributes being set through the `enrich` plugin in meteor. With these attributes, we plan to do some grouping of nodes/ filtering if necessary. 

I understand that this might be a very specific use case but given the user customizable attributes in this data field, I think this might be helpful in a broad sense also, adding props to nodes in lineage graph.

internal/store/postgres/asset_repository.go
bsushmith: Will check on this

core/asset/service.go
bsushmith: Given this - https://github.com/odpf/compass/pull/213#discussion_r1116556318

WDYT about the flag `with_node_props` ?

core/asset/service.go
bsushmith: Yes. Test cases are pending. will add this test case.

Issue #215, opened 2023-02-24
Getting asset's first version return no record for the asset error


Chief-Rishab:
**Describe the bug**
Getting a previous version of asset on the first version(`0.1`) is returning `no such record /:assetID` but works fine for other versions

**To Reproduce**
1. Create a new asset
2. Update the asset to create new version
3. Get asset's previous version with version id `0.1`

**Expected behavior**
Should return the first version of the asset

2FJDJD:
https://user-images.githubusercontent.com/128664730/229960176-a503b272-a891-4cea-80ed-ac68cff9aa55.mp4

https://github.com/orgs/odpf/projects/12/views/2?pane=issue&itemId=22062084

PR #221, opened 2023-04-03
fix: populate version in assets_versions for 1st entry

Chief-Rishab:
- Set version to 0.1 while inserting the first version of asset into
  assets_versions table.
- Update all rows in assets_versions to have 0.1 as the version if the
  column was not populated.

Closes: #215

### DIFF
--- a/internal/store/postgres/migrations/000016_update_assets_versions.up.sql
+++ b/internal/store/postgres/migrations/000016_update_assets_versions.up.sql
@@ -1,0 +1,1 @@
+UPDATE assets_versions SET version = '0.1' WHERE version = '';
--- a/internal/store/postgres/asset_repository.go
+++ b/internal/store/postgres/asset_repository.go
@@ -307,21 +307,22 @@ func (r *AssetRepository) Upsert(ctx context.Context, ns *namespace.Namespace, a
 
 	if fetchedAsset.ID == "" {
 		// insert flow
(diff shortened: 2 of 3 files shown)


internal/store/postgres/migrations/000016_update_assets_versions.up.sql
```diff
@@ -0,0 +1 @@
+UPDATE assets_versions SET version = '0.1' WHERE version = '';
```
kushsharma, line 1: Instead of modifying the db for it, can we handle it in go? Like if no version, assume 0.1?

Chief-Rishab: Getting the asset by version will be an issue in that case imo. Currently we're unable to get an asset's first version.

Chief-Rishab: In that case, instead of running migrations for version 0.1, I can add an extra case to query depending on the version number with:
```diff
+      if version == "0.1" {
+      ast, err := r.getByVersion(ctx, id, version, r.GetByID, sq.Eq{
+      sq.Eq{"a.urn": urn},
+      sq.IsNull{"a.version"},
+      } 
+       else{
	ast, err := r.getByVersion(ctx, id, version, r.GetByID, sq.Eq{
		"a.asset_id": id,
		"a.version":  version,
	})
+	}
```

Should we proceed with this?

kushsharma (approved):

PR #222, opened 2023-04-04
fix: allow creation of asset with owner identified by UUID

### LINKED ISSUE #216: Creating asset with uuid gives internal server error, owner info is empty in asset fetching requests
**Describe the bug**
Creating asset with owner's email is not getting reflected in the owner's field when we do a get request on the asset. (that is Asset fetching is returning empty owner's field) and

Creating asset with owner's uuid is giving an internal server error.

**To Reproduce**
1. Create an asset with body containing owner's information with these two cases
case 1: with owner email only
case 2: with owner's uuid ( will give an internal server error immediately )
2. `GET / {HOST}//v1beta1/assets` (for case 1)
3. See the owner's field for the asset

**Expected behavior**
Should contain owner details as defined during asset creation

Chief-Rishab:
- Fix create asset failure with owner(s) identified by UUID by fetching the Compass native user ID for the given UUID. Create the user if no user exists for the given UUID.
- Fix missing Owner's UUID field while creating asset patch
- Refactor SQL for inserting owners to use squirrel SQL builder.

Closes #216

### DIFF
--- a/internal/store/postgres/asset_repository.go
+++ b/internal/store/postgres/asset_repository.go
@@ -307,21 +307,22 @@ func (r *AssetRepository) Upsert(ctx context.Context, ns *namespace.Namespace, a
 
 	if fetchedAsset.ID == "" {
 		// insert flow
-		fetchedAsset.ID, err = r.insert(ctx, ns, ast)
+		id, err := r.insert(ctx, ns, ast)
 		if err != nil {
-			return fetchedAsset.ID, fmt.Errorf("error inserting asset to DB: %w", err)
+			return id, fmt.Errorf("error inserting asset to DB: %w", err)
 		}
-	} else {
-		// update flow
-		changelog, err := fetchedAsset.Diff(ast)
-		if err != nil {
-			return "", fmt.Errorf("error diffing two assets: %w", err)
-		}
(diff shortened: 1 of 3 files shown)


internal/store/postgres/asset_repository.go
```diff
@@ -633,31 +636,31 @@ func (r *AssetRepository) getOwners(ctx context.Context, assetID string) (owners
...
 
-	var values []string
-	var args = []interface{}{assetID}
-	for i, owner := range owners {
-		values = append(values, fmt.Sprintf("($1, $%d)", i+2))
-		args = append(args, owner.ID)
+	sqlb := sq.Insert("asset_owners").
+		Columns("asset_id", "user_id")
+	for _, o := range owners {
+		sqlb = sqlb.Values(assetID, o.ID)
```
kushsharma, line 651: Will this generate multiple `values` statements separated by commas?

Chief-Rishab: Yes, values statements with multiple comma seperated rows to be inserted in the table will be generated.
So in case the [] owners has two entries
```
VALUES (assetID, owners[0].ID), (assetID, owners[1].ID)
```

